PR #35, opened 2014-12-22
Add asynchronous value read/write capability on server side

paragonRobotics:
Asynchronous value reading/writing is required for our potential use of the node-opcua library, so this pull request is to add that functionality.  Variables can now optionally have a getAsync and setAsync function in order to asynchronously process requests.

However, this change required changes to a significant number of files, as the current architecture relied on "returns" and "getters/setters" which are all synchronous by nature.  This change affects the api for several calls, especially on the server side.  For example, server.write, server.writeSingleNode, server.readSingleNode, etc...all now require a callback instead of a synchronous return value.

All existing tests were updated to pass using the new API, and several new tests were added to test the functionality.

### DIFF
--- a/bin/simple_server.js
+++ b/bin/simple_server.js
@@ -119,6 +119,11 @@ server.on("post_initialize", function () {
                 get: function () {
                     var pump_speed = 200 + 100*Math.sin(Date.now()/10000);
                     return new Variant({dataType: DataType.Double, value: pump_speed});
+                },
+                getAsync: function(callback){
+                    var pump_speed = 200 + 100*Math.sin(Date.now()/10000);
+                    var test = new Variant({dataType: DataType.Double, value: pump_speed});
+                    callback(test);
                 }
             }
         });
--- a/lib/address_space/address_space.js
+++ b/lib/address_space/address_space.js
@@ -62,7 +62,7 @@ View.prototype.nodeClass = NodeClass.View;
  * @param attributeId
  * @return {DataValue}
  */
-View.prototype.readAttribute = function (attributeId) {
+View.prototype.readAttribute = function (attributeId, callback) {
 
     var options = {};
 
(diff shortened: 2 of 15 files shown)


paragonRobotics:
Sorry to ping again, but any thoughts on the asynchronous write functionality question?  The refreshFunc works great for the reads here, so we really just need the write functionality to get us to the point where we can use it for our needs.  Thanks so much--

erossignon:
Hi, thank you for your suggestion, this should be put in the wish list.
However I think that there is a way that this could be achieve now (waiting for a more comprehensive implementation.
Let me explain:
in the read operation , there is a need to fetch a fresh value of the variable asynchronously and wait for the callback to be called prior to returning the requested node value in the opcua request.
However, in the write operation we could imagine that the opcua write request simply initiates an asynchronous write operation and returns immediately without waiting for the asynchronous write op to be completed.

In this case the code would look like:

```javascript
  server.engine.addVariableInFolder(myDevices, {
        browseName: "Temperature",
        nodeId: "ns=2;s=Temperature",
        dataType: "Double",

        value: {
            refreshFunc: function (callback) {

                var temperature = 20 + 10 * Math.sin(Date.now() / 10000);
                var value = new Variant({dataType: DataType.Double, value: temperature});
                var sourceTimestamp = new Date();

                // simulate a asynchronous behaviour
                setTimeout(function () {
                    callback(null, value, sourceTimestamp);
                }, 100);
            }
            set: function(value) {
                  startAsyncWrite(value,function unused_callback(err,data){});
            }
        }
    });
```

paragonRobotics:
Thanks Etienne, however our requirements mandate that we get a confirmation from the wireless endpoint when the write was successfully received and processed (for critical control outputs).

Would you entertain the thought of adding a similar "refreshWriteFunc" capability to the OPCUA server:


// write services
OPCUAServer.prototype._on_WriteRequest = function (message, channel) {

    var server = this;

    var request = message.request;
    assert(request instanceof WriteRequest);
    assert(_.isArray(request.nodesToWrite));

    var results = [];

        assert(request.nodesToWrite[0]._schema.name === "WriteValue");
        results = server.engine.write(request.nodesToWrite);
        
    server.engine.refreshWriteValues(request.nodesToWrite,function(err,data){
         assert(_.isArray(results));
         assert(results.length === request.nodesToWrite.length);

	    var response = new WriteResponse({
		   results: results,
		   diagnosticInfos: null
	    });
	    channel.send_response("MSG", response, message);
    
    });
    
};


This refreshWriteFunc would also need to allow the return of response codes (like errors) to override the response code that was returned from the simple set function.  We would then relabel the existing refresh function as "refreshReadFunc" to separate them.  Please let me know your thoughts, and I would be willing to make the changes if you would like.  Thanks so much.

PR #71, opened 2015-04-28
some performance tuning

huazh:
Hi.

I want to use node-opcua to operate with some plc devices in low power computer. The functionality is well, but performance does not come up to expectation. So I fired up profiling and discovered that new Variant and Enum.get use too much cpu. These commits are based on the profiling.
The performance test I did is monitoring 1000 double variables with sampling interval 100ms and publishing interval 100ms. The original node-opcua ate up all cpu ( AMD E-450 1.6G Hz), but can not catch up the publishing interval ( got 90 publishing in 30s). After these fixes, the cpu usage is about 60% after JIT warm up, and no publishing is missing. It's about 4-5 times performance enhancement in pure monitoring application.
Hope these changes could be merged. All tests are passed on node v0.12.2.

### DIFF
--- a/lib/datamodel/access_level.js
+++ b/lib/datamodel/access_level.js
@@ -4,7 +4,7 @@
  */
 require("requirish")._(module);
 
-var Enum = require("enum");
+var Enum = require("lib/misc/enum");
 var assert = require("better-assert");
 
 var registerBasicType = require("lib/misc/factories_basic_type").registerBasicType;
--- a/lib/datamodel/nodeid.js
+++ b/lib/datamodel/nodeid.js
@@ -4,7 +4,7 @@
  * @module opcua.datamodel
  */
 require("requirish")._(module);
-var Enum = require("enum");
+var Enum = require("lib/misc/enum");
 var assert = require("better-assert");
 var isValidGuid = require("lib/datamodel/guid").isValidGuid;
 
--- a/lib/datamodel/numeric_range.js
+++ b/lib/datamodel/numeric_range.js
@@ -9,7 +9,7 @@ var StatusCodes = require("lib/datamodel/opcua_status_code").StatusCodes;
 
 var ec = require("lib/misc/encode_decode");
 var _ = require("underscore");
-var Enum = require("enum");
+var Enum = require("lib/misc/enum");
 var assert = require("better-assert");
 
 // OPC.UA Part 4 7.21 Numerical Range
--- a/lib/misc/encode_decode.js
+++ b/lib/misc/encode_decode.js
@@ -7,7 +7,7 @@
 require("requirish")._(module);
 
 var assert = require("better-assert");
-var Enum = require("enum");
+var Enum = require("lib/misc/enum");
 
 var ec = exports;
 var _ = require("underscore");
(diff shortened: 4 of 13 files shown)


erossignon:
Thank you for spotting this. The Enum component was working ok in version < 1.0.0 and has introduced this freezing stuff recently. I confoirme a x60 to x100 performance increase with your light implement :+1: 
I have opened a ticket to the Enum projet to let them know about this issue. https://github.com/adrai/enum/issues/21

huazh:
I don't think freezing is the actual problem. Enum creation only happened on initialization procedure, it should not slow down monitoring with freezing. The spot where light implement excels in is access with integer key, which node-opcua uses frequently in parsing packet. If I remove the enum creation from benchmark, there is no big difference in accessing enum with string key. But with integer, it shows 4-10x speedup (the last two).

  Benchmarking Enums
 CYCLE slowEnum op/s 35248.66541503015
 CYCLE fastEnum op/s 72447.2303206246
 Fastest is fastEnum
 Speed Up : x 2
    ✓ should verify that our enums are faster than old enums (1272ms)
 CYCLE slowEnum op/s 13190.49177278977
 CYCLE fastEnum op/s 36629.27950207223
 Fastest is fastEnum
 Speed Up : x 2
    ✓ should verify that our enums are faster than old enums (1123ms)
 CYCLE slowEnum op/s 31890.785152430213
 CYCLE fastEnum op/s 77190.20795213227
 Fastest is fastEnum
 Speed Up : x 2
    ✓ should verify that our enums are faster than old enums (1254ms)
 CYCLE slowEnum op/s 15869.851997760272
 CYCLE fastEnum op/s 65108.63662514895
 Fastest is fastEnum
 Speed Up : x 4
    ✓ should verify that our enums are faster than old enums (1190ms)
 CYCLE slowEnum op/s 6202.336347321556
 CYCLE fastEnum op/s 65322.549447467165
 Fastest is fastEnum
 Speed Up : x 10
    ✓ should verify that our enums are faster than old enums (1173ms)


huazh/node-opcua@1b6699ff69c7a7efda158c640448eb5813a9d6c1

huazh:
I benchmark againt Enum without freeze, it shows freeze actually slow down the access performance about 2 times. Without freeze, access with string key is the same as my implementation.
  Benchmarking Enums
 CYCLE slowEnum op/s 68920.94302841835
 CYCLE fastEnum op/s 74821.07251798527
 Fastest is fastEnum
 Speed Up : x 1
    ✓ should verify that our enums are faster than old enums (1346ms)
 CYCLE slowEnum op/s 30228.79254111068
 CYCLE fastEnum op/s 37207.53073862247
 Fastest is fastEnum
 Speed Up : x 1
    ✓ should verify that our enums are faster than old enums (1154ms)
 CYCLE slowEnum op/s 66678.59908263385
 CYCLE fastEnum op/s 89295.66085508122
 Fastest is fastEnum
 Speed Up : x 1
    ✓ should verify that our enums are faster than old enums (1334ms)
 CYCLE slowEnum op/s 24741.83418022731
 CYCLE fastEnum op/s 73496.99059233088
 Fastest is fastEnum
 Speed Up : x 2
    ✓ should verify that our enums are faster than old enums (1224ms)
 CYCLE slowEnum op/s 13665.065992739372
 CYCLE fastEnum op/s 71408.79419110095
 Fastest is fastEnum
 Speed Up : x 5
    ✓ should verify that our enums are faster than old enums (1191ms)

PR #84, opened 2015-06-12
Update client_session.js

demeyr:
Add ClientSession.prototype.readHistoryValue() function


erossignon:
We should add a dedicated new unit test for this new feature. Test could go in test/end_to_end.
This would involved a little bit more work as we need to implement a rudimentary read counterpart method  on the server. Would you fancy trying this ?

demeyr:
Maybe take a look next week.

PR #90, opened 2015-07-09
Events

hanskhe:
Hello,

This PR implements basic event support on the client. If EventNotifier is specified for monitoring, then it will instantiate the correct filter classes. It will throw an error if filter is in the object when not needed. The filter supported is EventFilter without a whereClause.

Correct schemas for ```createMonitoredItemResponse``` are implemented. 

Monitoring parameters needed to create a monitor is on the following format:
````
monitoringParameters: {
    samplingInterval: 0, 
    discardOldest: true,
    queueSize: 10,
    filter: {
        extensionType: 'EventFilter',
        data: {
            selectClauses: [
                [{namespaceIndex:0, name: "NodeId"}]
            ],
            whereClause: []
        }
    }
}
````

As you know, the filter is an extension type, so we have created a data format to specify what filter type it is. Do you think this is a good idea or should it only accept correctly instantiated filters?

We are also missing tests at the moment. For that we need a basic server implementation for event monitoring.

We are at the moment unsure whether we should indicate an error with the event emitter, or by throwing exceptions. The important thing is that monitor creation is stopped, as certain servers will crash if the parameters are not correct.

### DIFF
--- a/bin/crypto_create_CA.js
+++ b/bin/crypto_create_CA.js
@@ -658,7 +658,7 @@ function create_default_certificates(done) {
 
 function __create_default_certificates(base_name,prefix,application_URI,done) {
 
-    // Fugl hack
+    // Bad hack that ensures that paths with spaces are correctly interpreted.
     base_name = "\"" + base_name + "\"";
 
     assert(_.isFunction(done));
--- a/bin/generate_opcua_classes.js
+++ b/bin/generate_opcua_classes.js
@@ -242,9 +242,9 @@ registerObject("ReadProcessedDetails");
 registerObject("ReadAtTimeDetails");
 
 // Event results
-registerObject("ContentFilterElementResult", undefined, true);
-registerObject("ContentFilterResult", undefined, true);
-registerObject("EventFilterResult", undefined, true);
+registerObject("ContentFilterElementResult");
+registerObject("ContentFilterResult");
+registerObject("EventFilterResult");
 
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by erossignon** [comment]
Thank you for your contribution. after reviewing the code I think that we can simplify the creation of the filter object by using something along those line:

```
 filter: new subscription_service.EventFilter({
                selectClauses: [// SimpleAttributeOperand
                    {
                        typeId: "i=123", // NodeId

                        browsePath: [    // QualifiedName
                            {namespaceIndex: 1,name:"A"},{namespaceIndex: 1,name:"B"},{namespaceIndex: 1,name:"C"}
                        ],
                        attributeId: AttributeIds.Value,
                        indexRange: new NumericRange()
                    },
                    {
                        // etc...
                    },
                    {
                        // etc...
                    }
                ],
                whereClause: { //ContentFilter
                    elements: [ // ContentFilterElement
                        {
                            filterOperator: subscription_service.FilterOperator.IsNull,
                            filterOperands: [ //
                                new subscription_service.ElementOperand({
                                    index: 123
                                }),
                                new subscription_service.AttributeOperand({
                                    nodeId: "i=10",
                                    alias: "someText",
                                    browsePath: { //RelativePath

                                    },
                                    attributeId:  AttributeIds.Value
                                })
                            ]
                        }
                    ]
                }
            })
        });
```

PR #97, opened 2015-07-27
Initial crawl write crawl test

perok:
Hi,

We have started to write a test for our crawler problem. It's crudely based on your tests.

The problem is that the crawler does not call the success callback after a write operation (tested on node 'EnabledFlag' (n=0;i=2294)).

The idea is to test that the nodecrawler returns data after a write operation. Currently we are trying to write to 'EnabledFlag' (n=0;i=2294). We get the error ''The node id refers to a node that does not exist in the server address space." on the write operation. Could you guide us in the right direction?

Thanks

### DIFF
--- a/test/end_to_end/test_opcua_ClientServer_crawl_write_crawl.js
+++ b/test/end_to_end/test_opcua_ClientServer_crawl_write_crawl.js
@@ -1,0 +1,150 @@
+
+require("requirish")._(module);
+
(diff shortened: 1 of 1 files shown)


erossignon:
Hi
It looks like the node i=2294 (ServerDiagnostics.EnabledFlag) is not present in the server we use for testing. Let me add a few missing bits.

PR #153, opened 2016-01-06
Create monitor.js

pelis:
process monitored items, reconnect, when connection breaks


erossignon:
@pelis ; would it be better to submit your example files in a gist rather than as a pull-request ?

pelis:
OK will do

PR #207, opened 2016-05-14
If used timestamped_set, no context is set

GermanBluefox:
Bind to variable if timestamped_set used to determine context


GermanBluefox:
Hello Etienne,
thank you for very useful module!
Do you understand right the pull request? 
I can't start the production version without this fix. It would be nice if you can merge it.

thank you,
Bluefox

erossignon:
Please also provide a unit test to stress this feature and make sure there is no regression in the future.

Issue #292, opened 2017-01-04
Moving toward ES6


erossignon:
I think we should really move toward ES6 for readility of code and use of Promise as well and other nice stuff (import, modules ) 
This would mean dropping support for old version of nodejs. 
Please vote with a 👍 if you don't care switching on node 6 or above or 👎 if you would rather not now.

zenonpelle:
Hi!I
My vote is 👍

/Per
Från: Etienne [mailto:notifications@github.com]
Skickat: den 4 januari 2017 17:43
Till: node-opcua/node-opcua <node-opcua@noreply.github.com>
Kopia: Subscribed <subscribed@noreply.github.com>
Ämne: [node-opcua/node-opcua] Moving toward ES6 (#292)


I think we should really move toward ES6 for readility of code and use of Promise as well and other nice stuff (import, modules )
This would mean dropping support for old version of nodejs.
Please vote with a 👍 if you don't care switching on node 6 or above or 👎 if you would rather not now.

BLamy:
I would love to see this. Since the architecture is very OOP it may be worth looking into typescript shouldn't be that much more effort because most data types are already defined in the docs. 

If going with plain ES6 I would highly suggest using AirBNBs eslint rules. I've had very good success with upgrading ES5 code to ES6 using those rules.

gilesbradshaw:
I think it's a great idea to switch - however do you have to drop old versions of node?  Pretty sure you'd be able to compile your ES6 to ES5 with babel..

typescript is a great idea too..

gilesbradshaw:
I've got quite far doing an ES6 conversion - it would be good if you had a look  https://github.com/ua-ql/node-opcua

It doesn't build and install automatically quite properly yet but it will soon.  

I've web packed it to one bundle so pretty sure it will run on previous versions of node + wold be possible to use  ES2016+ and typescript etc.

There are a few things that needed changing - mainly do do with dynamic requires (you cant do things like require(avariable) as it sucks in everything as a possible match (ie all the test code) also this means you can't build and then require stuff 'on the fly' (you probably *could* using hot reloading) so I've split 'createObject) from registerObject and in places you have circular dependencies and things like exporting created schemas from tests which then use those schemas.

However I have got the code through the tests (modified to cope with webpack/import restrictions and also run a server and a client with it...

[...]

PR #295, opened 2017-01-09
ES6

BLamy:
Here are some suggestions for getting started with es6.

- [x] Add a yarn.lock
- [x] use airbnb's eslint rules
- [x] transpile with babel es2015 preset
- [x] use babel module resolver instead of requireish
- [ ] use [lebab](https://github.com/lebab/lebab) if possible


BLamy:
Even if the library never moves to ES6 I still think the babel module resolver should be used over requireish. requireish has burned me several times. 

image

test/transport/test_client_tcp_transport.js
```diff
@@ -177,7 +177,7 @@ describe("testing ClientTCP_transport", function () {
         });
         fake_server.pushResponse(spyOnServerWrite);
 
-        transport.timeout = 10; // very short timeout;
+        transport.timeout = 100; // very short timeout;
```
BLamy, line 180: The comment makes me think there was a reason it was only at 10.

gilesbradshaw: you might well be right - I think @erossignon wld need to look at that..

gilesbradshaw: ah ok - now passing on Node 6+  :)

BLamy:
@gilesbradshaw 

A little confused as to why node 5 completely crapped out. 
Node 4 is just missing a polyfill. 

.12 Issues can probably be resolved but I think an argument could be made for dropping it all together.

gilesbradshaw:
I've been playing with this and have ES6ed opcua-client and client_base and then started lebabing the various transforms.  However it seems to be the tests are not completely determinate.  I now get errors even when I go back to the revision I have PRed to you....

gilesbradshaw:
Not sure if tests are totally robust - I am now getting #298 even on original...

BLamy:
I'm going to hold off on merging your [giant PR](https://github.com/BLamy/node-opcua/pull/3) until we hear from @erossignon. I want to make sure he agrees with the opinions we've taken thus far before I overwhelm him with one giant PR.

gilesbradshaw:
no problem - good idea!

erossignon:
Hi, I am investigating how we could bring @gilesbradshaw's ES6 branch that is now compiling OK into a 'ES6-experimental branch" of the official node-opcua repository , so we could keep going together .
Not quite sure yet, what is the best route: Giles has forked Bret's fork, so I cannot merge Giles' PR directly I guess.

BLamy:
Looks like the es6 branch got published.

Closing in favor of tracking progress here:
https://github.com/node-opcua/node-opcua/issues/292

PR #300, opened 2017-01-25
began formatting to ESLint rules

BLamy:


lib/address_space/add-event-type/install.js
```diff
@@ -5,57 +5,58 @@
 
 import assert from "better-assert";
 import _ from "underscore";
+import { randomBytes } from "crypto";
```
BLamy, line 8: Was there a reason these were imported inside of the function? I pulled them out.

erossignon: no, no reason for those one. you can put them at the top of the file

lib/address_space/add-enumeration-type/install.js
```diff
@@ -27,7 +24,8 @@ function install(AddressSpace) {
      * @param options.enumeration[].value       {Number}
      * @param options.enumeration[].description {String|LocalizedText|null}
      */
-  AddressSpace.prototype.addEnumerationType = function (options) {
+    /* eslint-disable max-len */
```
erossignon, line 27: Why change indent here ? AddressSpace.prototype.addEnumerationType  is inside the install function and shall be indented?

lib/address_space/alarms_and_conditions/ConditionSnapshot.js
```diff
@@ -560,35 +555,35 @@ function prepare_date(sourceTimestamp) {
...
+  // structure for the ConditionVariable Value Attribute.
 
   assert(node.typeDefinitionObj.browseName.toString() === "ConditionVariableType");
-  assert(node.sourceTimestamp.browseName.toString() == "SourceTimestamp");
+  assert(node.sourceTimestamp.browseName.toString() === "SourceTimestamp");
   node.on("value_changed", _update_sourceTimestamp);
 }
 
 
-import { BrowseDirection } from "lib/services/browse_service";
```
erossignon: Be carefull some import have to stay inside the install function to break cyclic dependencies

erossignon:
obviously, tests must pass for PR to be accepted.

BLamy:
I'll take a look at why the test are failing tonight.

gilesbradshaw:
Ah sorry didn't see this I think I have caused all sorts of conflicts with it - sorry 

I have also just done an eslint fix with npm run fix which has put all intentation to 4 spaces

gilesbradshaw:
good to have you involved though @BLamy i've been all over the code so possibly causing all sports of conflict with stuff others are doing..

gilesbradshaw:
Also I don't *think* importing in a code block will work - especially with webpack - where necessary due to cyclic dependencies i left it as a require..

PR #317, opened 2017-03-07
implemented browseNext in client session

xhronos:
The continuation points returned by client session browse() can be passed to browseNext() to get the next chunk of results.


erossignon:
Hi Xhronos, browseNext  has now been implemented, thank you for your PR that was inspiring

PR #347, opened 2017-05-02
Add/Fix validateDataType() for EnumType

tomsoftware:
If the value of an EnumType is written by the UA Client the Server raises: 
EXCEPTION CAUGHT WHILE PROCESSING REQUEST !!! WriteRequest datatype 2:MyEnumeration must be registered
Error: datatype 2:MyEnumeration must be registered


erossignon:
Please add a dedicated unit test to capture and demonstrate the behavior

PR #371, opened 2017-09-06
Update typescript declarations

bompi88:
Updates the following declarations:
- [x] `ClientMonitoredItem` extends `EventTarget`. Can be used in RxJs.
- [x] defines ´on´in ´ClientSubscription´


erossignon:
ClientSubscription  now has on method declaration

PR #373, opened 2017-09-15
added missing type information for: StatusCode,  DataType, Variant and VariantArrayType

JulZimmermann:


JulZimmermann:
Unfortunately we can't use the current version of node-opcua in our typescript project.
This is because type information for functions and types we use are not defined in the d.ts files.
I added the type information we need but it's still not complete.

I'd love to see this merged

PR #385, opened 2017-10-25
Fill BrowseResponse interface definition

lochbrunner:
Fills the empty interface `BrowseResponse` and fixes the typo when using them in the `browse` method.

PS. I am not sure what type `nodeId` and `nodeClass` have. I have got different object literals there.


erossignon:
In fact , this has been fixed already ... we can drop it , thank you for contributing.

PR #387, opened 2017-10-27
fixed #386 client crash on terminating monitored item before initialization

### LINKED ISSUE #386: Terminating un-initialized monitored item crashes client
**I'm submitting a ...**  (check one with "x")

 - ( x ) bug report => search github for a similar issue or PR before submitting
 - ( ? ) feature request
 - ( ) support request => Please do not submit support request here, instead see use 
       [gitter](https://gitter.im/node-opcua/node-opcua) or [stackoverflow](https://stackoverflow.com/questions/tagged/node-opcua)

**Current behavior**
Calling terminate on a monitored item before the initialized callback has fired causes the client to crash.

**Expected behavior**
Calling terminate on a monitored item should terminate the monitored item.

**Minimal reproduction of the problem with instructions**

```
var monitoredItem = the_subscription.monitor(
    {
        nodeId: "ns=1;s=PumpSpeed",
        attributeId: AttributeIds.Value
    },
    {
        samplingInterval: 250,
        //xx filter:  { parameterTypeId: "ns=0;i=0",  encodingMask: 0 },
        queueSize: 10000,
        discardOldest: true
    }
);

// the following line crashes the client
monitoredItem.terminate();

monitoredItem.on("initialized", function () {
    console.log("monitoredItem initialized");
    // terminate call is okay here, but crashes client if called before initialized
    //callback();
});
```

  
**What is the motivation / use case for changing the behavior?**
It is possible for an action to occur which would trigger terminating a monitored item immediately after calling subscription.monitor() for that item. In my application the node-opcua client wrapped up into a web server, and this behavior would allow a client of my web server to crash it by subscribing and then immediately unsubscribing an item. 

**Please tell us about your environment:**

  * ( ) I have installed node-opcua from source ( using git clone)
  * ( x ) I have installed node-opcua as a package ( using npm install )
  * ( ) I am using an application that uses node-opcua 
  
       - ( ) node-red
       - ( ) other : please specify 
        
  * Device: _____
  * OS version: _____
    
    - ( ) Windows : version : _________
    - ( ) Linux   : version : _________
    - ( x ) MacOs   : version : 10.13
    - ( ) Raspbian: version : _________
    - ( ) Other   : specify :
   
   * Description of the other OPCUA system I am trying to connect to:
   
     - Name: Sample Client / Server from node-opcua-samples
     - Version:_________
     - Manufacturer/Software vendor:_________
     - link : https://
     
    
* **node-opcua version:** ``
Installed from npm
    "node-opcua": "0.1.0-11",
    "node-opcua-samples": "0.1.0-11"

* **Node:** 
   `node --version` = v6.11.1

I would be happy to provide a pull request implementing a potential fix for this issue if you'd consider a merge.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/50783682-terminating-un-initialized-monitored-item-crashes-client?utm_campaign=plugin&utm_content=tracker%2F526033&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F526033&utm_medium=issues&utm_source=github).
</bountysource-plugin>

JoshMcguigan:
I have tested that this does fix #386. I don't see many tests specifically for the OPCUA client, but the ones that are there still pass. Let me know if you'd like to see additional test coverage for this change.

### DIFF
--- a/packages/node-opcua-client/src/client_subscription.js
+++ b/packages/node-opcua-client/src/client_subscription.js
@@ -646,9 +646,17 @@ ClientSubscription.prototype._remove = function (monitoredItem) {
     var self = this;
     var clientHandle = monitoredItem.monitoringParameters.clientHandle;
     assert(clientHandle);
-    assert(self.monitoredItems.hasOwnProperty(clientHandle));
-    monitoredItem.removeAllListeners();
-    delete self.monitoredItems[clientHandle];
+
+    var deleteMonitoredItem = function(){
+        monitoredItem.removeAllListeners();
+        delete self.monitoredItems[clientHandle];
+    };
+
+    if(self.monitoredItems.hasOwnProperty(clientHandle)){
+        deleteMonitoredItem();
+    } else {
+        monitoredItem.on("initialized", deleteMonitoredItem);
+    }
 };
 
 ClientSubscription.prototype._delete_monitored_items = function (monitoredItems, callback) {


erossignon:
@JoshMcguigan : a dedicated unit test would be welcome to ensure that behavior is as expected and that we do not introduce regression further on.

PR #394, opened 2017-11-11
node-opcua-nodeid - Adjust NodeId creation on coerceNodeId

gbaquedano:
Hello, one of my applications detected the following behaviour when creating a NodeId.
A NodeId instantiation from coerceNodeId could end on an assertion error (no identifierType).

### DIFF
--- a/packages/node-opcua-nodeid/src/nodeid.js
+++ b/packages/node-opcua-nodeid/src/nodeid.js
@@ -270,9 +270,9 @@ function coerceNodeId(value, namespace) {
         console.log( "xxxx VALUE = ",value);
 
         var tmp = value;
-        value = value || tmp.value;
-        namespace = namespace || tmp.namespace;
-        identifierType = identifierType || tmp.identifierType;
+        value = tmp.value || value;
+        namespace = tmp.namespace || namespace;
+        identifierType = tmp.identifierType || identifierType;
         return new NodeId(identifierType, value, namespace);
     }
     return new NodeId(identifierType, value, namespace);
--- a/packages/node-opcua-nodeid/src/nodeid.js
+++ b/packages/node-opcua-nodeid/src/nodeid.js
@@ -270,10 +270,10 @@ function coerceNodeId(value, namespace) {
         console.log( "xxxx VALUE = ",value);
 
         var tmp = value;
-        value = tmp.value;
+        value = value || tmp.value;
         namespace = namespace || tmp.namespace;
-        identifierType = tmp.identifierTypes;
-        return new NodeId(value, namespace);
+        identifierType = identifierType || tmp.identifierType;
+        return new NodeId(identifierType, value, namespace);
     }
     return new NodeId(identifierType, value, namespace);
 }


biancode:
same problem in Node-RED while testing node-opcua 0.2.0

biancode:
Please, I need that fix for Node-RED

marcelwinh:
I also encountered this problem when I updated from 0.0.65 to 0.2.1. However, this problem only occurs with some variables in my code, so I've examined it further.
I think i found the problem in my code where I'm creating the variables.
So i tested it with the sample_server.js here it is with my workaround:

This works:
```
addressSpace.addVariable({
	componentOf: device,
	browseName: "MyVariable2opcua.DataType.Double.key",
	dataType: opcua.DataType.Double.key,	// <=== workaround
	value: {
		get: function () {
			return new opcua.Variant({dataType: opcua.DataType.Double, value: 123 });
		}
	}
});
```

This doesn't work anymore (it worked in 0.0.65):		
```
addressSpace.addVariable({
	componentOf: device,
	browseName: "MyVariable3opcua.DataType.Double",
	dataType: opcua.DataType.Double,
	value: {
		get: function () {
			return new opcua.Variant({dataType: opcua.DataType.Double, value: 123 });
		}
	}
});
```

After i found a workaround for me i didn't investigated it further. Except your pull request, it also solves the problem.

The sample_server way of creating a variable is:
```
addressSpace.addVariable({
	componentOf: device,
	browseName: "MyVariable1",
	dataType: "Double",			// <== which is the same as 'opcua.DataType.Double.key'
	value: {
		get: function () {
			return new opcua.Variant({dataType: opcua.DataType.Double, value: variable1 });
		}
	}
});
```

So i don't know if it was a bug <0.1.0 that this worked or it is now a bug that it doesn't work anymore.

erossignon:
this issue could happen when node-opcua is installed from source and if two versions of node-opcua are coexisting in the same environement.
Usually this could be sorted out by  deleting node_modules and package-index.json and redoing ```npm install```. I haven't come up to a easy way to detect conflicting version easily yet

erossignon:
I am reluctant to change "dataType: opcua.DataType.Double" into dataType: opcua.DataType.Double.key as this is ugly and simply would have to be propagated everywhere. I would rather tackle the source cause of the problem : i.E double inclusing of two conflicting node-opcua-constants modules.

marcelwinh:
Just to clarify i don't think this is a bug. It works like you have described in sample_server:
```
addressSpace.addVariable({
	componentOf: device,
	browseName: "MyVariable1",
	dataType: "Double",
	value: {
		get: function () {
			return new opcua.Variant({dataType: opcua.DataType.Double, value: variable1 });
		}
	}
});
```
for the variable node generation the dataType parameter is specified as string (an enumaration of som strings "Double", "String", "Boolean", ...) and not as type "opcua.DataType" like in the variant.

But in version 0.0.65 you could also give it as opcua.DataType. Here is the question was it a bug or a feature.

```
addressSpace.addVariable({
	componentOf: device,
	browseName: "MyVariable1",
	dataType: "Double" || opcua.DataType.Double , // bug or a feature?
	value: {
		get: function () {
			return new opcua.Variant({dataType: opcua.DataType.Double, value: variable1 });
		}
	}
});
```
I wouldn't change "dataType: opcua.DataType.Double" into dataType: opcua.DataType.Double.key too. It was just meant as tip or quick & ugly workaround if someone has this error.

PR #397, opened 2017-11-21
Elegant null handling in string utils

mboorstin:
I had node-opcua crash when crawling a server because the crawler passed a null `referenceType.browseName.name` to `lowerFirstLetter` ([node_crawler.js:1009](https://github.com/node-opcua/node-opcua/blob/5d3446d7befee48fb55f8a2416c8b0ef2ac3b282/packages/node-opcua-client-crawler/src/node_crawler.js#L1009)).  I'm not sure whether `lowerFirstLetter` should be handling this more elegantly or whether the null `referenceType.browseName.name` is evidence of a deeper problem in the crawler; presuming it's the former I've updated `lowerFirstLetter` (and its companion `capitalizeFirstLetter`) to elegantly handle nulls.  I've added appropriate tests (and renamed an existing test file in the same directory to reflect the fact that it's just testing `buffer_ellipsis`).  Please let me know if this looks good, or if there's a deeper problem in the crawler that I should be fixing instead.  Thanks!


mboorstin:
Note that a few of the tests are failing on this PR's build, but they look like they're inconsistently failing on other PR's as well so I don't think they're from my changes.

erossignon:
Not providing a valid browseName for a reference seems to be a error of the server you're trying to connect to. Can you describe which server you've been trying to connect to to get this bug appearing ?
(Nevertheless, I agree that node-opcua should be robust enough to cope with inconsistencies on the other end)

mboorstin:
@erossignon we're trying to connect to the onboard OPC/UA server on a Wago 750-8101 PLC ([product page here](http://www.wago.us/products/components-for-automation/modular-io-system-series-750-753/programmable-fieldbus-controller/pfc-family/)).  Unfortunately the PLC is on a customer's network, so I can't give you a public IP address to connect to, but I'm happy to run any diagnostics that would be helpful for you.

mboorstin:
@erossignon since you mentioned above that node-opcua should be robust enough to cope with inconsistencies on the other end, are there any changes you would like me to make to this pull request so that it can be merged?

PR #405, opened 2017-12-01
fix: getCurrentUserRole

everhardt:
`getCurrentUserRole` in SessionContext looks at `this.session.userManager`, but that never exists. I think it should be `this.server.userManager`, but that isn't always populated. Therefore, this pull:

* lets `getCurrentUserRole` look at `this.server.userManager`
* ensures `SessionContext` is always constructed with a `server` option
* also (minor), I've added yarn.lock to the gitignore file, as I've used yarn instead of npm

Note: we could even add an asset to the `SessionContext` constructor that `options.server` should be defined, but then the tests also need to be rewritten. It would be more safe though, as user authentication depends on it


erossignon:
Please provide a comprehensive unit test to validate your code so it can be merged

everhardt:
The test I've committed is an adaptation of the test in `packages/node-opcua-end2end-test/test/end_to_end/test_e2e_issue405_control_user_role.js`. The existing one verifies that an anonymous user can't get read and write access when he was not allowed to, but it missed the case where a user with the operator operator should not get access to something only an admin should have access to. 

The test I've committed therefore fails if you run it before my fix (and passes now).

erossignon:
@everhardt , can you check that this is now ok on node-opcua@0.2.1 ?

everhardt:
Yes, it is, thank you!

PR #406, opened 2017-12-06
fix: stringToQualifiedName to split browsename after ':' only if the first part is a number

marcelwinh:
If you have a browsename like 'device:ip' this function splits it into
ns='device' & browseName='ip'. I have added another check so this
happens only if the first part is a number.


erossignon:
We cannot merge without a proper unit test. Can you provide one ?

marcelwinh:
I have extended the function to pass the test. I hope i haven't forgotten anything.

PR #451, opened 2018-03-07
support security Modes/Policies with settings None/None and Username&pw

zombinary:
changes to connect a server with following secure settings

> node simple_client  -e opc.tcp://localhost:49230 -P=None -s=NONE -u=testUser -p=""

For the securityPolicy in the **getCryptoFactory()** i copied the global variabel **_Basic128Rsa15** and renamed it _**None** and changed the securityPolicy in **SecurityPolicy.None**.
This works for me but i'm not sure that the other settings return the correct value in my new global vairable **_None**.

### DIFF
--- a/packages/node-opcua-client/src/opcua_client.js
+++ b/packages/node-opcua-client/src/opcua_client.js
@@ -380,16 +380,21 @@ function createUserNameIdentityToken(session, userName, password) {
     var userIdentityToken = new UserNameIdentityToken({
         userName: userName,
         password: Buffer.from(password, "utf-8"),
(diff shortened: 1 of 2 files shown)


zombinary:
i wanted to ask about the state from this request. 
Is it possible to take the changes in to the master trunk or is it conceivable when this secure mode is solved?

erossignon:
Hi @zombinary, I cannot accept this PR as : 
 - it doesn't explain how it complies with the OPCUA specification
 - there is no unit test to demonstrate what this PR is fixing.

erossignon:
@zombinary , the issue that was resolved here will be handled differently in #574.

Issue #467, opened 2018-03-26
Failed to change event filter on EventNotifierObject


bruegth:
**I'm submitting a ...**  (check one with "x")

 - (x) bug report => search github for a similar issue or PR before submitting
 - ( ) feature request

**I will make an pull request to fix that behavior, too.**

**Current behavior**

When changing EventFilter the method ModifyMonitoredItemsRequest will failed because of AssertionError:
```
EXCEPTION CAUGHT WHILE PROCESSING REQUEST !! ModifyMonitoredItemsRequest
{ /*ModifyMonitoredItemsRequest*/
 requestHeader                 /* RequestHeader                    */: {
   authenticationToken         /* NodeId                           */: ns=0;b=56e5537810df2a386d3a292d60171622
   timeStamp                   /* UtcTime                          */: 2018-03-23T12:37:15.736Z
   requestHandle               /* IntegerId                        */: 1011191               0xf6df7
   returnDiagnostics           /* UInt32                           */: 0               0x0
   auditEntryId                /* UAString                         */: null
   timeoutHint                 /* UInt32                           */: 10000               0x2710
   additionalHeader            /* ExtensionObject                  */: null
 }
 subscriptionId                /* IntegerId                        */: 414511               0x6532f
 timestampsToReturn            /* TimestampsToReturn               */: Both ( 2)
 itemsToModify                 /* MonitoredItemModifyRequest    [] */: [
   { /*0*/
     monitoredItemId           /* IntegerId                        */: 2               0x2
     requestedParameters       /* MonitoringParameters             */: {
       clientHandle            /* IntegerId                        */: 10               0xa
       samplingInterval        /* Duration                         */: 500
       filter                  /* ExtensionObject                  */: { /*EventFilter*/
 selectClauses                 /* SimpleAttributeOperand        [] */: [
   { /*0*/
     typeId                    /* NodeId                           */: ns=0;i=2041
     browsePath                /* QualifiedName                 [] */: [
       { /*0*/
         namespaceIndex        /* UInt16                           */: 0
         name                  /* String                           */: EventId
       }
     ]
     attributeId               /* IntegerId                        */: 13               0xd
     indexRange                /* NumericRange                     */: NumericRange:<Empty>
   },
   { /*1*/
     typeId                    /* NodeId                           */: ns=0;i=2041
     browsePath                /* QualifiedName                 [] */: [
       { /*0*/
         namespaceIndex        /* UInt16                           */: 0
[...]

PR #468, opened 2018-03-26
adds minimumSamplingInterval to UAObject

bruegth:
This will close Issue #467


erossignon:
@bruegth  thank you for your pull request and the effort you put into it
I am sorry that this fixes the symptom only and not the acutal problem.
UAObject do not have minimumSamplingInterval attribute as per specification
image
I am sorry to reject this pull request based on this,

PR #469, opened 2018-03-26
adds displayname to object- and variable-types

bruegth:
adds displayname to object-/variabletypes.
maybe change default behavior if no displayname is present to something else maybe empty?


erossignon:
Please add one or more unit tests to demonstrate/test the behavior and provide example of use.

bruegth:
Okay, I will do that, also for my other pull request with method displayName.

Current behavior is that a browsename with namespace (e.g. "2:EventNotifierObject" or "2:GetData") will result in a displayname which contains the namespace and the BrowseName is not a localized Text.

It will take a litte time, because of Eastertime.

PR #472, opened 2018-04-04
Add missing dependencies to node-opcua

ipa-mdl:
a3a3d2d4d51b18e95ba427d572b8cffc2c0b55dc required ` semver`  and `colors`.
While `colors` gets pulled in by the sub-packages, ` semver` gets not (anymore?).


DimitrijeManic:
@erossignon npm should be updated with this fix :+1:

PR #486, opened 2018-05-23
fixed BadFilterNotAllowed on boolean, String, ByteString & LocalizedText

marcelwinh:
Server returns BadFilterNotAllowed if the DataChange Trigger is changed on a non numeric Monitored Item with DeadbandType = None.
It should only return BadFilterNotAllowed when the DeadbandType != None


erossignon:
Thank you for your fix and comprehensive unit test to demonstrate how and why this fix is needed.

PR #508, opened 2018-07-25
better displayName

beppemarazzi:
if displayName is not specified now use browseName AFTER it is decoded by _get_QualifiedBrowseName (eventually removes namespaceindex prefix). i.e: browseName = "1:MyName" now displayName is "MyName", before was "1:MyName"

### DIFF
--- a/packages/node-opcua-address-space/src/base_node.js
+++ b/packages/node-opcua-address-space/src/base_node.js
@@ -115,7 +115,7 @@ function BaseNode(options) {
     this.browseName = _get_QualifiedBrowseName(options.browseName);
 
     // re-use browseName as displayName if displayName is missing
-    options.displayName = options.displayName || options.browseName.name.toString();
+    options.displayName = options.displayName || this.browseName.name;
 
     this._setDisplayName(options.displayName);
 


beppemarazzi:
Thank you for your work!!!! I'm out for some holidays. Next week i'll provide...

beppemarazzi:
i'm trying to reproduce the bad behavior and i concluded that my modification had sense when (in old versions) was possible to specify namespace index with "colon syntax" i.e. addressSpace.addFolder(addressSpace.rootFolder.objects, "2:FooBar");

beppemarazzi:
Other info: refer to #513

1. versions <= 0.4.0 allowed to create a node with colon in browseName (i.e. browseName: "1:MyObject")

2. version 0.4.2 throws if you try to create a node with colon in browseName (i.e. browseName: "1:MyObject")

3. after 0d004de browseName with colons is deprecated but allowed

I'm sure that bad behavior (getting a displayName with colon) fixed by this PR brings out in case 1 and i guess it can be also in case 3. ASAP i'll try to reproduce it with a unit test

beppemarazzi:
Just tested with master. Now if you specify a colon in browseName, a warning is issued but all the name specified (comprising prefix and colon) is used for the text part and the prefix is no more used as namespace index.
So this makes usefulness this PR

Issue #504, opened 2018-07-17
Connection cannot be established from 3rd party embedded client to node-opcua-server


nospam2000:
**I'm submitting a ...**  (check one with "x")

 - (x) bug report => search github for a similar issue or PR before submitting
 - ( ) feature request
 - ( ) support request => Please do not submit support request here, instead see use 
      - use [gitter](https://gitter.im/node-opcua/node-opcua) or [stackoverflow](https://stackoverflow.com/questions/tagged/node-opcua) for community support, or 
      - [contact Sterfive for commercial support](https://www.sterfive.com)

**Current behavior**

A 3rd party embedded device OPC UA client drops the TCP connection right after it receives the OPC UA ACK answer to its HEL request from node-opcua-server.
The ReceiveBufferSize and the SendBufferSize of the ACK response are larger than the values which are requested by the client (8196 vs. 8192 according to a Wireshark capture).

### The HEL request from the 3rd party client:
OpcUa Binary Protocol
    Message Type: HEL
    Chunk Type: F
    Message Size: 60
    Version: 0
    **ReceiveBufferSize: 8192
    SendBufferSize: 8192**
    MaxMessageSize: 2097152
    MaxChunkCount: 256
    EndPointUrl: opc.tcp://141.73.114.71:8080

### The Ack response from the node-opcua-server:
OpcUa Binary Protocol
    Message Type: ACK
    Chunk Type: F
    Message Size: 28
    Version: 0
    **ReceiveBufferSize: 8196
    SendBufferSize: 8196**
    MaxMessageSize: 2097152
    MaxChunkCount: 256

**Expected behavior**

The ReceiveBufferSize and the SendBufferSize of the response should be smaller than or equal to the values which are requested by the client (8192 instead of 8196).

**Minimal reproduction of the problem with instructions**

  1. use a client with a ReceiveBufferSize and SendBufferSize of 8192 byte
  2. connect to a node-opcua-server
  3. verify the values of ReceiveBufferSize and SendBufferSize in the ACK response which is sent to the client. They have to be less than or equal to the values requested by the client.

  
**What is the motivation / use case for changing the behavior?**

I wish to be able to connect from the 3rd party client to the node-opcua-server.

**Please tell us about your environment:**

The server operating system is Windows 10 build 1803
"node-opcua-server"   version "0.4.1" (from npmjs.org)

  * ( ) I have installed node-opcua from source ( using git clone)
  * (x ) I have installed node-opcua as a package ( using npm install )
[...]

PR #524, opened 2018-09-17
changed check for minBufSize to 8192 for secured connections

nospam2000:
and to 64 bytes for non-secured connections. Fixes issue #504.

I'm totally sure, that the minsize of 8196 is incorrect and has to be 8192 (see my comments in the issue #504).
I'm not sure if the error handling I added is correct.

### DIFF
--- a/packages/node-opcua-secure-channel/src/message_chunker.js
+++ b/packages/node-opcua-secure-channel/src/message_chunker.js
@@ -64,7 +64,7 @@ MessageChunker.prototype.update = function(options) {
  * @param msgType {String}
  * @param options
  * @param options.tokenId
- * @param options.chunkSize    [default=8196]
+ * @param options.chunkSize    [default=8192]
  *
  * @param options.signatureLength  {Integer} [default=0]
  * @param options.signingFunc {Function} [default=null]
--- a/packages/node-opcua-secure-channel/src/secure_message_chunk_manager.js
+++ b/packages/node-opcua-secure-channel/src/secure_message_chunk_manager.js
@@ -29,7 +29,7 @@ exports.chooseSecurityHeader = chooseSecurityHeader;
  *
  * @param msgType
  * @param options
- * @param options.chunkSize {Integer} [=8196]
+ * @param options.chunkSize {Integer} [=8192]
  * @param options.secureChannelId
  * @param options.requestId
  * @param options.signatureLength  {Integer}  [undefined]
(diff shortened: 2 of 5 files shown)


erossignon:
It would be nice if you could provide a dedicated unit test

nospam2000:
> It would be nice if you could provide a dedicated unit test

Can you tell me where to start from? I haven't found any other tests in the "test" folder, just an empty file.

I get the following error when I try to install the latest version from the master branch during the "npm install" command:
```
/Users/michael/Downloads/node-opcua-buffer-fix/node_modules/yargs/lib/command.js:22
    globalMiddleware.push(...middlewares)
                          ^^^

SyntaxError: Unexpected token ...
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/Users/michael/Downloads/node-opcua-buffer-fix/node_modules/yargs/lib/argsert.js:2:17)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
```

nospam2000:
I upgraded to node10 and the install problem is gone.

The question is still: Can you tell me where to start from, is there already a test which can be used as base? I need a secured communication and a non-secured communication test case.

Describing the tests is easy:
1. connect with unsecured connection, receiveBufferSize = 512  => expect OK
2. connect with unsecured connection, receiveBufferSize = 8192  => expect OK
3. connect with secured connection, receiveBufferSize = 512  => expect FAIL
4. connect with secured connection, receiveBufferSize = 8192  => expect OK
5..8 the same with sendBufferSize

The buffer sizes in client_tcp_transport.js:ClientTCP_transport.prototype._send_HELLO_request() are hardcoded and not accessibe from outside, I have to look how it can be set. Maybe it requires an interface extension for this.

  Michael

erossignon:
Michael,
Note that the specifications state that buffer size shall be at least 8192 bytes, 

image

erossignon:
@nospam2000 Michael, do you want to adjust this PR by simply setting 8192 as minimum ? 
I believe you deserve to have your commit merged to node-opcua, once adapted with the findings of our last discussion, However If  you run out of time on your end I am happy to do it on my side.
your choice .. let me know.
( I'd like to incorporate a fix for this issue into next node-opcua release)

nospam2000:
@erossignon 
> do you want to adjust this PR by simply setting 8192 as minimum ?
> However If you run out of time on your end I am happy to do it on my side.

Please do that for me, next week I won't have any time.

  Michael

PR #547, opened 2018-10-21
Removed obsolete Security Policies

cacamille3:
Removed references to Basic128, Basic192, Basic192Rsa15, Basic256Rsa15
Reduced Basic256 Security Policy level from 3 to 0 (same as
Basic128Rsa15)


packages/node-opcua-server/test/test_server_endpoint.js
```diff
@@ -143,23 +135,6 @@ describe("OPCUAServerEndpoint#getEndpointDescription", function () {
         server_endpoint = null;
     });
 
-
-    it_with_crypto("should not find a endpoint matching MessageSecurityMode.SIGN and SecurityPolicy.Basic128", function () {
```
erossignon, line 147: Please do not simply remove this test ... please adapt it with valid policy instead

packages/node-opcua-secure-channel/src/security_policy.js
```diff
@@ -65,12 +65,8 @@ const _ = require("underscore");
 const SecurityPolicy = new Enum({
     Invalid: "invalid",
     None: "http://opcfoundation.org/UA/SecurityPolicy#None",
-    Basic128: "http://opcfoundation.org/UA/SecurityPolicy#Basic128",
```
erossignon, line 68: For testing and validation purposes , it would be nice to find a way to keep those obsolete enum value so we can write functional tests that validates for instance than third parties clients and server act as request by the 1.0.4 version of the specification toward old security policies

packages/node-opcua-end2end-test/test/secure_connection/test_e2e_secure_all_possibe_secure_connection.js
```diff
@@ -414,14 +414,6 @@ function perform_collection_of_test_with_client_configuration(message, options)
         common_test("Basic256", "SIGNANDENCRYPT", options, done);
     });
 
-    it("should fail    with Basic256Rsa15  with Sign           " + message, function (done) {
```
erossignon, line 417: Please adapt; do not remove

cacamille3:
I don't think that this is a good idea to have obsolete Security Policies defined in the code. 
First there is no more references to any of thoses obsolete security policies in the OPC UA specification or Stack written by the OPC Foundation.
Second, this is confusing for other packages using node-opcua. node-red-contrib-opcua for instance was still having UI component referencing Basic128: #https://github.com/mikakaraila/node-red-contrib-opcua/pull/90)
The only possibility to still have some attended failing unit test will be to use "Invalid" instead of those obsolete Security Policies.

erossignon:
>> I don't think that this is a good idea to have obsolete Security Policies defined in the code.
Agreed,  ! I was suggesting however there must be a way to specify those "out of date" policy for testing and conformance testing purposes.

This being said, tests that validates that existing/missing policies are handled properly .and should not be wiped out but adapted.

PR #591, opened 2019-02-08
readVariableValue() should return both timestamps

ovives:
I think it could be a good option to return the `serverTimestamp` and the `sourceTimestamp` when the client reads the value of a variable. 
Is interesting to be able to know when the signal was recorded and when it was stored in the server especially if you are working with time/value pairs in tiny granularities.


erossignon:
readVariableValue is ment to read ... variable value only, please use read(...) to read all the other related attribute

Issue #607, opened 2019-03-07
Crash when using node-opcua server inside electron on Mac and Linux


ptillemans:
**I'm submitting a ...**  (check one with "x")

 - (x) bug report => search github for a similar issue or PR before submitting
 - ( ) feature request
 - ( ) support request => Please do not submit support request here, instead see use 
      - use [gitter](https://gitter.im/node-opcua/node-opcua) or [stackoverflow](https://stackoverflow.com/questions/tagged/node-opcua) for community support, or 
      - [contact Sterfive for commercial support](https://www.sterfive.com)

**Current behavior**

When importing *opcua-server* in an electron app.

```
import { startOpcuaServer } from './opcua/opcua-server';
```

It causes the program to never end importing dependencies and go into an endless loop of starting new electron applications till memory runs out.

**Expected behavior**

App continues with main routine after importing and display the UI. This is for the UI of an industrial control application which is running on equipment on the shopfloor.

**Minimal reproduction of the problem with instructions**

  1. create a minimal electron app. e.g. https://github.com/electron-react-boilerplate/electron-react-boilerplate
  2. add a dependency on opcua : yarn add node-opcua
  3. and an import : import { startOpcuaServer } from './opcua/opcua-server';
  4. package the app for linux : yarn package-linux
  5. install the package: sudo dpkg -i release/electron-react-boilerplate_0.17.1_amd64.deb     
  6. run the app : /opt/ElectronReact/electron-react-boilerplate 
  7. observe that the app is hanging and no output is generated.
  8. open another terminal and run *top* and observe the electron processes breed like rabbits.
  
What happens is that the *fqdn* module used by node-opcua-hostname package depends on an old version of shelljs. Shelljs uses *process.execPath* to find the name of the *node* executable to run small dynamically generated javascript scripts to run system commands. In most cases the *node* executably is the top level process and this works fine. For electron apps, this actually points to electron app which wraps node and starts a browser and the 2 javascript enviroments for the frontend and backend. 

When the fqdn library is used during imports it, it tries to call *hostname* by this method by executing the script with call to the program pointed to by *process.execPath* mistakensy assuming it points to the node interpreter. This keeps starting the electron app and initializing the app till memory runs out.

The fqdn library and shelljs libraries are only used from node-opcua-hostname package, and can be relatively easy be replaced with *os.hostname()* plus a dns lookup on Mac and Linux.

[...]

PR #608, opened 2019-03-07
alternative fqdn implementation without fqdn+shelljs deps

ptillemans:
see https://github.com/node-opcua/node-opcua/issues/607

this is an implementation without relying on fqdn and the old version of shelljs.

Tested to work on linux, mac, including debian packaged for linux.

It essentially does the same as fqdn, but just using the standard node library.


fbalazsy:
Merge it please !

ptillemans:
Seems no longer necesary. Thanks.

PR #613, opened 2019-04-09
Check certificate extensions for null before creating the URI

patricia-dm:
If the certificate being checked for creation of the URI is X509 v1, then tbsCertificate.extensions is null, meaning that there's no defined value to apply to applicationUri. This fix simply checks whether the extensions property is null (aka this is a v1 certificate) before trying to pull data from children of the extensions. It also allows the library to be compiled with strictNullChecks enabled for typescript.


erossignon:
this pr has been merged manually and is available in node-opcua@2.0.0-alpha.22

PR #659, opened 2019-08-13
Added ListOfString to the nodeset loader.

Magnulas:
The function for parsing ListOfString xml elements was missing from the nodeset loader. ListOfString elements were loaded as just a single string instead of a list.


erossignon:
Can you provide a unit test so the PR can be incorporated ?

erossignon:
Great ! Thank you very much ...

PR #667, opened 2019-08-28
Use ClientSession instead of ClientSessionImpl in node-opcue-client-proxy

eivindArundo:
After upgrading to 2.1.5 in my project, `tsc` started complaining about code in `node_modules/node-opcua-client/source/private`, and I got similar errors when trying to build `node-opcua`. 

This seems to fix it, hope it's useful.


erossignon:
Thank you for your contribution. This change make sense .

gautejohan:
Thank you for acting so quick! 

Do you know when you are planning the next release (with this change)?

PR #695, opened 2019-11-05
Update tcp_transport.ts

gilesbradshaw:
The code is raising an error event which in my app is ending up setting _socket to null.
Therefore the code is unsafe.

This is not the most elegant way to fix this but it illustrates the problem.


erossignon:
Hi Gil,
I think I found the same one as well and it has now been addressed in master.
I hope you'll find the time to test it once node-opcua@2.3.0 is ok and confirm this.

erossignon:
Hi Giles thank you for your contribution,  I guess we can now close this PR  .

PR #698, opened 2019-11-11
Remove self referencing import

mrothNET:


erossignon:
Hi I found this one as well when switching to latest tsc version. It is already incorporated in master now and will be available in next release.
Thank for your PR and contribution.

PR #756, opened 2020-04-09
AccessLevel and UserAccessLevel not added to xml

DaniHaag:
When the nodeset is exported to xml the accesslevel and the useraccesslevel are not taken into account


erossignon:
Hi DaniHaag,
Thank you for your Pull request.
We would need to have unit tests fix and may be one more unit tests added so it can be merged.
If you need help let me know ...
Etienne

erossignon:
similar change added by 9b40b9a0e9fad727c2c00fe029166db84a03a95d ( including a unit test)
Thank you for your contribution

PR #766, opened 2020-04-22
Prevent implicit 'any[]' typing

Razzeee:
I'm not 100% sure why our project picks up this problem but your linter settings do not. But I feel like this is the best way to report this.


Razzeee:
This might have something to do with the inclusion of the `*.ts` files in the package that node-opcua ships? As `--skipLibCheck` might have been an option, but it only skips `*.d.ts` files.

erossignon:
OK, this is now fixed in node-opcua. Thank you for spotted this out

Razzeee:
First version including this fix is `v2.6.0-alpha.0`?

PR #767, opened 2020-04-26
fix overloaded variable bug

erossignon:


erossignon:
fix overloaded variable bug for #767

PR #784, opened 2020-05-13
Bugfixes in simple-client-ts.ts and simple-client.js

pirminkager:
- added require("util")
    - namespace array fix


packages/node-opcua-samples/bin/simple_client.js
```diff
@@ -444,7 +445,7 @@ async function main() {
     console.log(" --- NAMESPACE ARRAY ---");
     const namespaceArray = dataValue.value.value;
     for (const namespace of namespaceArray) {
-        console.log(" Namespace ", namespace.index, "  : ", namespace);
+        console.log(" Namespace ", namespaceArray.indexOf(namespace), "  : ", namespace);
```
erossignon, line 448: Please revert, the intend is really  `namespace.index`

packages/node-opcua-samples/bin/simple_client_ts.ts
```diff
@@ -455,7 +455,7 @@ async function main() {
     console.log(" --- NAMESPACE ARRAY ---");
     const namespaceArray = dataValue.value.value;
     for (const namespace of namespaceArray) {
-        console.log(" Namespace ", namespace.index, "  : ", namespace);
+        console.log(" Namespace ", namespaceArray.indexOf(namespace), "  : ", namespace);
```
erossignon, line 458: Please revert, the intend is really  `namespace.index`

pirminkager: I am getting `undefined` when i call `namespace.index` but i would expect the Indexnumber.
Why could that be?

erossignon: sorry,I confirmed ,  you are right.
 namespaceArray returns a bunch of string not a array of Namespace Object.

Issue #647, opened 2019-07-18
OPCUA Server does not start with alternateHostname


Camarbro:
**I'm submitting a ...**  (check one with "x")

 - (X) bug report => search github for a similar issue or PR before submitting
 - ( ) feature request
 - ( ) support request => Please do not submit support request here, instead see use 
      - use [gitter](https://gitter.im/node-opcua/node-opcua) or [stackoverflow](https://stackoverflow.com/questions/tagged/node-opcua) for community support, or 
      - [contact Sterfive for commercial support](https://www.sterfive.com)

**Current behavior**

When creating an OPC UA server and specifying in the server options the alternate hostname as a single option or as part of alternate endpoints, this option gets ignored and uses the computer's hostname.

EDIT:
After debugging for a while I noticed that if a specify the alternate hostname the server will duplicate the endpoints, once for the hostname of the computer and once for the alternate hostname, and when connecting to the server an error appears: 

`22:01:27.490Z :server_end_point              :140   Several matching endpoints have been found : `
`22:01:27.490Z :server_end_point              :142       opc.tcp://computerhostname:4335/Sample None http://opcfoundation.org/UA/SecurityPolicy#None`
`22:01:27.491Z :server_end_point              :142       opc.tcp://alternateHostname:4335/Sample None http://opcfoundation.org/UA/SecurityPolicy#None`

Example:
`
const server = new opcua.OPCUAServer({
    resourcePath: "Sample",
    port: 4335,
    alternateHostname: "someotherhostname"     
});
// When started, the server opcua address looks loke opc.tcp://computerHostname:4335/Sample
`

**Expected behavior**

`
const server = new opcua.OPCUAServer({
    resourcePath: "Sample",
    port: 4335,
    alternateHostname: "someotherhostname"     
});
// Expected server with hostname opc.tcp://someotherHostname:4335/Sample
`

**Minimal reproduction of the problem with instructions**

Use the example code provided above and create a simple server
  1. ________________
  2. ________________
  3. ________________
  ...
  
**What is the motivation / use case for changing the behavior?**

Previously in version 0.7.0, this would work perfectly but since I update to version 2.1.0 it doesn't

**Please tell us about your environment:**

Node opcua server in Mojave

  * ( ) my request is related to node-opcua acting as a  OPCUA CLIENT 
  * ( X)  my request is related to node-opcua acting as a OPCUA SERVER 

  * ( ) I have installed node-opcua from source ( using git clone)
[...]

PR #822, opened 2020-06-30
Expose endpoint hostname option in server options

padraic-padraic:
Related to issue #647, this is a fairly lightweight change to expose the hostname option when creating the default endpoints.

### DIFF
--- a/packages/node-opcua-server/source/opcua_server.ts
+++ b/packages/node-opcua-server/source/opcua_server.ts
@@ -11,7 +11,7 @@ import { EventEmitter } from "events";
 import * as _ from "underscore";
 import { callbackify } from "util";
 
-import { extractFullyQualifiedDomainName } from "node-opcua-hostname";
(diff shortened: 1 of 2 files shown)


packages/node-opcua-server/source/opcua_server.ts
```diff
@@ -3313,7 +3320,8 @@ export class OPCUAServer extends OPCUABaseServer {
     if (!endpointOptions) {
       throw new Error("internal error");
     }
-
+    var hostname = getFullyQualifiedDomainName();
```
erossignon, line 3323: please use const for consistancy

erossignon: It would be nice to have a dedicated unit test as well

erossignon:
@padraic-padraic  thank you for the proposed PR ! I have done a quick code-review and have added 2 comments. 
Let me know if you need assistance or help to setup the requested unit test.

PR #827, opened 2020-07-10
feat: add WebSocket Server Support (opcua+uacp)

demike:
- add  ws protocol support
- add  wss protocol support 
  - works with the generated certificates
  - reuses the same certificate / key (mechanism) for wss already provided in the library  
- add a tcp2ws proxy for e2e Testing
- add e2e tests for ws and wss (using the tcp2ws proxy)
- add support for describe.only / describeWithLeakDetector.only


demike:
> : {array: [], statusCode: StatusCode } { =>
> I beleve that array could be a [] or a Buffer
> 
> so return value should be
> : {array: any[] | Buffer, statusCode: StatusCode } { =>
> 
> we could also try generic
> 
> public set_values(arrayToAlter: ArrayOrBuffer, newValues: any) : {array: ArrayOrBuffer, statusCode: StatusCode } {

@erossignon  Of course you are right! Sorry I totally missed this comment!

erossignon:
Closing due to inactivity,

PR #828, opened 2020-07-13
Fix #826: Change included fields in Create Session response

### LINKED ISSUE #826: server.applicationUri not returned in CreateSessionResponse
**Current behavior**

While investigating a communication problem between the node-opcua server and the NI OPC UA client, we noticed in the messages that the server endpoints in the CreateSessionResponse do not contain the server.applicationUri. The OPC-UA spec mentions that server.applicationUri should be amongst the information retained in the CreateSessionResponse (https://reference.opcfoundation.org/v104/Core/docs/Part4/5.6.2/#5.6.2.2)

**Describe the bug**
The server does not return all required information in

**To Reproduce**
1. Create an opcua server following e.g. an example
2. Enabled debugging by setting the environment variable NODEOPCUADEBUG=SERVERTRACE,TRACE,PERF
3 Inspect the server endpoints during a connection attempt

**Expected behavior**
The CreateSessionResponse includes the server applicationUri

**Screenshots**

**Gist**
gist: https://gist.github.com/padraic-padraic/0b90069fbcf551c66dac36ef33c39058

 * ( ) my request is related to node-opcua acting as a  OPCUA CLIENT 
  * (x)  my request is related to node-opcua acting as a OPCUA SERVER 

  * ( ) I have installed node-opcua from source ( using git clone)
  * (x) I have installed node-opcua as a package ( using npm install )
  * ( ) I am using an application that uses node-opcua 
  
       - ( ) node-red
       - ( ) other : please specify 
        
  * Device: AWS EC2 server
  * OS version: 
    
    - ( ) Windows : version : _________
    - (x) Linux   : version : Debian Buster, Ubuntu 18.04 LTS
    - ( ) MacOs   : version : _________
    - ( ) Raspbian: version : _________
    - ( ) Other   : specify :
   
   * Description of the other OPCUA system I am trying to connect to:
   
     - Name:_________
     - Version:_________
     - Manufacturer/Software vendor:_________
     - link : https://
     

* **node-opcua version:**  : 2.7.0

* **Node:** 
   `node --version` = v14.2.0

padraic-padraic:


erossignon:
@padraic-padraic, great stuff,
Can we also update the comment (that came from the 1.03 part 4) specification:
```
  // It is recommended that Servers only include the endpointUrl, securityMode,
  // securityPolicyUri, userIdentityTokens, transportProfileUri and securityLevel with all
  // other parameters set to null. Only the recommended parameters shall be verified by
  // the client.
```
with the new phrase in the 1.04 version
image

>> It is recommended that Servers only include the **server.applicationUri,** endpointUrl, securityMode, securityPolicyUri, userIdentityTokens, transportProfileUri and securityLevel with all other parameters set to null. Only the recommended
parameters shall be verified by the client.


Also, we could also add a new unit test or adjust an existing one, to specifically verify that the server: { applicationUri } is added and all other parameters of the server property set to null.

packages/node-opcua-server/source/opcua_server.ts
```diff
@@ -281,9 +281,15 @@ function getRequiredEndpointInfo(endpoint: EndpointDescription) {
     securityPolicyUri: endpoint.securityPolicyUri,
     transportProfileUri: endpoint.transportProfileUri,
     userIdentityTokens: endpoint.userIdentityTokens
+    server: {applicationUri: endpoint.server.applicationUri}
   });
   // reduce even further by explicitly setting unwanted members to null
-  (e as any).server = null;
+  (e as any).productUri = null;
```
erossignon, line 254: I don't think this is needed to set those to null ? 

  (e as any).productUri = null;
  (e as any).applicationName = null;
  (e as any).applicationType = null;
  (e as any).gatewayServerUri = null;
  (e as any).discoveryProfileUri = null;
  (e as any).discoveryUrls = null;


Reading the spec it seems that only  server: {applicationUri: endpoint.server.applicationUri} was missing.

erossignon:
merging this change is now handled in #891

PR #908, opened 2020-11-03
Define repository line endings

MarcoSeb:
added .gitattributes file and changed line endings as described here:
https://docs.github.com/en/free-pro-team@latest/github/using-git/configuring-git-to-handle-line-endings


MarcoSeb:
In my opinion a .gitattribute file is helpfull to have deterministic line endings. It makes it independent of the configuration of each developer. If you dont agrree just abonden this PR

erossignon:
Hi Seb, I am not sure I can accept this one as it changing too many file and create a lot of noise.
I think it needs to stick with the convention already in place in node-opcua rather than switching to other conventions that might be in place on your side.

MarcoSeb:
okay

PR #929, opened 2021-01-10
Add support for boolean data in coerceVariantType

### LINKED ISSUE #794: coerceVariantType on string "true"
Hi,

I am pulling values from a CSV file and running coerceVariantType() on the value before writing to the server.

Works for all except, booleans.

`
   let valueToWrite = coerceVariantType(DataType.Boolean, "true")
`

This will return a string, not true. Any support for that in the future ?

maneetgoyal:
Fixes #794


packages/node-opcua-variant/test/test_variant.js
```diff
@@ -1460,6 +1460,10 @@ describe("miscellaneous Variant tests", () => {
         coerceVariantType(DataType.Int32, { value: 26, key: "foo" }).should.eql(26);
     });
 
+    it("coerceVariantType on Boolean", () => {
+        coerceVariantType(DataType.Boolean, true).should.eql(true);
```
erossignon, line 1464: Please add a test to check that "true"|"false" can be converted to true|false.

 coerceVariantType(DataType.Boolean, "true" /* a string here  */).should.eql(true);
 
coerceVariantType(DataType.Boolean, "false" /* a string here  */).should.eql(false);

erossignon (changes requested):
Please add a unit test to verify that the suggested change produces the v expected result.

packages/node-opcua-variant/test/test_variant.js
maneetgoyal: On it 👍

maneetgoyal: Done, please check.

Issue #627, opened 2019-05-30
Can't connect to a server that uses only secured endpoints


kevinherron:
The node-opcua sample client can't connect to server that is configured as follows:

An endpoint at `opc.tcp://localhost:62541/milo/discovery` that implements only Discovery services and allows connections with no security.

Endpoints at `opc.tcp://localhost:62541/milo` that require security, no exceptions, including the discovery services.

This configuration is outlined in Part 6, Section 7.4, where it defines "/discovery" appended to the base URL as a well-known endpoint to be used when a server requires a different address from the unsecured Discovery services.

When the node-opcua client calls GetEndpoints on the discovery endpoint, the endpoint URLs in the returned `EndointDescriptions` correctly point to the `opc.tcp://localhost:62541/milo`, but the client does not follow those and instead tries to create a session against the discovery endpoint.

You can download a demo server to test against [here](https://www.dropbox.com/s/xc5gv22t3aojvkq/milo-demo-server-0.1-SNAPSHOT.zip?dl=0).

Unzip it, `cd` into the directory, edit `config/server.json` and remove "None" from the `securityPolicyList`.

Run `chmod +x bin/server.sh` and then `bin/server.sh`. You need to have JDK 11 installed and the `JAVA_HOME` environment variable pointing to it.

erossignon:
Thank you for spotting this:
![image](https://user-images.githubusercontent.com/680220/59186566-76ac1600-8b73-11e9-9dfc-49b82dd35315.png)

We shall adjust the way we access server endpoints when server only exposes secure end points and a a discovery endpoint.

PR #933, opened 2021-01-17
avoid exposure of restricted endpoints in createsessionresponse message

treestructure:
filter out restricted endpoints from the create session response. In the _on_CreateSessionRequest, the endpoint to use in the session needs to  be an non-restricted endpoint. Some OPC clients e.g. UA Expert then try to connect using an UserIdentityToken that is not present for the endpoint and cannot be found in the function:

findUserTokenByPolicy of opc_server.ts


erossignon:
Hi Martin,
thank you for this PR;.
We use TDD. 
It would be nice if you could elaborate a "non-regression" unit test for this enhancement prior we can merge the PR. Let me know if you need help.

erossignon:
May be related to #627

PR #965, opened 2021-03-19
Bugfix Client Sample

rickypid:


erossignon:
@rickypid , thank you for providing this fix. The sample file is actually auto generated by processing the corresponding md file . We needed to fix the md file as well, this is now done.

Issue #985, opened 2021-04-23
Server crashes upon automatic client reconnection with bad sessionid


paolodenti:
OPCUAServer crashes with the message below

```
21:08:38.628Z :base_server                   :464    messages  EXCEPTION CAUGHT WHILE PROCESSING REQUEST !!! ActivateSessionRequest
 ...                                                Attempt to access memory outside buffer bounds
 ...                                                RangeError [ERR_BUFFER_OUT_OF_BOUNDS]: Attempt to access memory outside buffer bounds
 ...                                                    at boundsError (internal/buffer.js:81:11)
 ...                                                    at Buffer.readUInt32LE (internal/buffer.js:217:5)
 ...                                                    at OPCUAServer.userNameIdentityTokenAuthenticateUser (<my folder>/node_modules/node-opcua-server/dist/opcua_server.js:880:33)
 ...                                                    at OPCUAServer.isUserAuthorized (<my folder>/node_modules/node-opcua-server/dist/opcua_server.js:930:25)
 ...                                                    at <my folder>/node_modules/node-opcua-server/dist/opcua_server.js:1306:20
 ...                                                    at OPCUAServer.isValidUserNameIdentityToken (<my folder>/node_modules/node-opcua-server/dist/opcua_server.js:795:16)
 ...                                                    at OPCUAServer.isValidUserIdentityToken (<my folder>/node_modules/node-opcua-server/dist/opcua_server.js:907:25)
 ...                                                    at OPCUAServer._on_ActivateSessionRequest (<my folder>/node_modules/node-opcua-server/dist/opcua_server.js:1295:16)
 ...                                                    at OPCUAServer.on_request (<my folder>/node_modules/node-opcua-server/dist/base_server.js:288:25)
 ...                                                    at OPCUAServerEndPoint.<anonymous> (<my folder>/node_modules/node-opcua-server/dist/opcua_server.js:503:26)
21:08:43.873Z :base_server                   :301   [NODE-OPCUA-W08] EXCEPTION CAUGHT WHILE PROCESSING REQUEST !! ActivateSessionRequest
21:08:43.873Z :base_server                   :302   { /*ActivateSessionRequest*/
 ...                                                 requestHeader                 /* RequestHeader       */: {
 ...                                                   authenticationToken         /* NodeId              */: ns=0;b=94c1addc65b7c5a5ce5ec8e5d57aa15e
 ...                                                   timestamp                   /* DateTime            */: 2021-04-23T21:08:41.555Z
[...]

PR #986, opened 2021-04-24
Fix for issue #985

paolodenti:
When an invalid certificate is received, the server replies with failed authentication instead of crashing on a bad memory address on `buff.readUInt32LE(0)`


erossignon:
Hi @paolodenti , thank you for your PR, It occurs that I also fixed the same issue, by using a slightly different technique,

PR #1008, opened 2021-05-27
emit Error object if MonitoredItemGroup fails on creation 

treestructure:
if the createMonitoredItems call inside the  MonitoredItemToolbox returns an error ( e.g. BadTooManyOperations) from the server, this error cannot be handled specifically at the moment because the terminated event is thrown without any error information so the user cannot handle the situation specifically.

### DIFF
--- a/packages/node-opcua-client/source/client_monitored_item_group.ts
+++ b/packages/node-opcua-client/source/client_monitored_item_group.ts
@@ -15,7 +15,7 @@ export interface ClientMonitoredItemGroup extends EventEmitter, ClientMonitoredI
 
     on(event: "err", eventHandler: (message: string) => void): this;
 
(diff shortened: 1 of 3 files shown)


erossignon:
Thank you;
Can you provide a unit test ?

treestructure:
What would be the best spot to create a unit test  ? Currenty I did not find a unit test covering the behaviour of MonitoredItemGroup or ClientMonitoredItemToolbox classes. 

Would `packages/node-opcua-client/test` be the right spot ?

treestructure:
There is now a unittest in packages/node-opcua-client/test/test_monitoreditemgroup.ts that checks if the error gets passed to the user of the MonitoredItemGroup class

erossignon (approved):

PR #1023, opened 2021-07-02
Added custom hostname support to the OPCUA Discovery Server

XOlifreX:
In the project I work on currently we have a LDS deployed on a Kubernetes cluster in the cloud. When trying to register a server on it, every time the LDS would take its local hostname, a name that cannot be resolved by any server who's on a different network. 

This pull request adds a way to define custom hostnames to the OPCUADiscoveryServer class in the same manner the OPCUAServer can. Now you can pass a custom hostname that does get resolved outside of the network.

### DIFF
--- a/packages/node-opcua-server-discovery/source/opcua_discovery_server.ts
+++ b/packages/node-opcua-server-discovery/source/opcua_discovery_server.ts
@@ -14,7 +14,7 @@ import { makeApplicationUrn } from "node-opcua-common";
 import { checkDebugFlag, make_debugLog } from "node-opcua-debug";
 import { extractFullyQualifiedDomainName, resolveFullyQualifiedDomainName } from "node-opcua-hostname";
 import { Message, Response, ServerSecureChannelLayer } from "node-opcua-secure-channel";
-import { OPCUABaseServer, OPCUABaseServerOptions, OPCUAServerEndPoint } from "node-opcua-server";
+import { OPCUABaseServer, OPCUAServerEndpointOptions, OPCUABaseServerOptions, OPCUAServerEndPoint } from "node-opcua-server";
 
 import {
     Announcement,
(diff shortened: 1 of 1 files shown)


erossignon:
ALthough I understand the need to be able to access the LocalDiscovery server under different hostname.  I don't think it is appropriate to use the OPCUAServerEndpointOptions  interface to convey, this information; OPCUAServerEndpointOptions  is used for fullblown servers and contains options that are irrelevant for LDS only servers. 

We will need to find a more subtil approach.

packages/node-opcua-server-discovery/source/opcua_discovery_server.ts
```diff
@@ -135,7 +145,15 @@ export class OPCUADiscoveryServer extends OPCUABaseServer {
                 privateKey: this.getPrivateKey(),
                 serverInfo: this.serverInfo
             });
-            endPoint.addStandardEndpointDescriptions();
+            
+            endPoint.addStandardEndpointDescriptions({
+                securityModes: this._primaryEndpoint.securityModes,
+                securityPolicies: this._primaryEndpoint.securityPolicies,
+                disableDiscovery: this._primaryEndpoint.disableDiscovery,
```
erossignon, line 152: Not sure disableDiscovery is relevant for a ... discovery server

packages/node-opcua-server-discovery/source/opcua_discovery_server.ts
```diff
@@ -55,9 +55,8 @@ function hasCapabilities(serverCapabilities: UAString[] | null, serverCapability
     return !!serverCapabilities.join(" ").match(serverCapabilityFilter);
 }
 
-export interface OPCUADiscoveryServerOptions extends OPCUABaseServerOptions {
+export interface OPCUADiscoveryServerOptions extends OPCUABaseServerOptions, OPCUAServerEndpointOptions {
```
erossignon, line 58: I would rather add a àlthername hostname property to the OPCUADiscoveryServerOptions  interface

OPCUAServerEndpointOptions  contains too many options that are not relevant to LDS server.

packages/node-opcua-server-discovery/source/opcua_discovery_server.ts
```diff
@@ -113,6 +113,16 @@ export class OPCUADiscoveryServer extends OPCUABaseServer {
 
         super(options);
 
+        this._primaryEndpoint = {
```
erossignon, line 116: super.options already contains a copy of the options parameter.
No need to duplicate here .

erossignon (changes requested):
Pleas also add relevant units test to verify that the alternate hostnames are properly exposed and present in the auto-generated certificate.

erossignon:
@XOlifreX, shall we close this PR that look staled ? , may be you found a suitable workaround.

Issue #1060, opened 2021-10-05
Read of attribute AccessLevelEx is not possible.


ata1704:
If I do a session.read of any nodeId with the attributeId "AccessLevelEx" (27) I get the following error:
_invalid value 27 for field attributeId of type UInt32_

1.  I'm using the following code:
`const readResult = await session.read({
nodeId: nodeId,
attributeId: AttributeIds[attributeId]
});`
2. Normally I'm getting a result and check for a bad status code of readResult to know if the attribute is available in the node and do my own error handling if not.
3. When I use "AccessLevelEx" as the attributeId the read function throws an error no matter what nodeId I use.

**Context**

- (x) my request is related to node-opcua acting as an OPCUA CLIENT
- ( ) my request is related to node-opcua acting as an OPCUA SERVER

- ( ) I have installed node-opcua from source ( using git clone)
- (x) I have installed node-opcua as a package ( using npm install )
- ( ) I am using an application that uses node-opcua

- Device: Macbook Pro (15-inch, 2018)
- OS version: 11.2.3

- Description of the other OPCUA system I am trying to connect to:

    - Name: Prosys OPC UA Simulation Server
    - Version: 5.0.08-330
    - Manufacturer/Software vendor: Prosys
    - link to the manufacturer or product site: https://www.prosysopc.com/

- **node-opcua version:** :
  node-opcua@2.42.0

- **Node:**
  `node --version` = v14.15.5

PR #1064, opened 2021-10-07
fix for #1060

AndreasHeine:
fix for "UnhandledPromiseRejectionWarning: Error:  invalid value 27 for field attributeId of type UInt32"


AndreasHeine:
@erossignon somehow i dont see you review... anyways have added the test as you requested 😉

Issue #1093, opened 2021-12-02
Input argument diagnostics always returned and requestHeader not available in MethodFunctor arguments


hoancea:
**Current behavior**

When binding a function to a UAMethod, the `requestHeader` cannot be found inside any of the arguments of the function. The `requestHeader` contains a flag that specifies if the input argument diagnostics should be returned or not, according to the link below.
https://reference.opcfoundation.org/v104/Core/docs/Amendment3/5.7.3/

**Describe the bug**

At the moment, the diagnostics are always returned, regardless of what is in the request header. 

**Step by step instruction to reproduce**

1. use `bindMethod` of `UAMethod`
2. debug inside of the bound method
3. `requestHeader` not available anywhere

**Expected behavior**

Either provide us with the `requestHeader` inside the `SessionContext` in order to add the diagnostics ourselves OR strip away the returned diagnostic information if the flag (`returnDiagnostics`) on the `requestHeader` is set to true.

**Screenshots**

**Gist**

 
gist:

**Context**

- (X) my request is related to node-opcua acting as an OPCUA SERVER
- (X) I have installed node-opcua as a package ( using npm install )

- **node-opcua version:** : 2.59.0
  
  

- **Node:**
  `node --version` = 12.22.1

erossignon:
The specs says
![image](https://user-images.githubusercontent.com/680220/144423109-94b6872d-6ef0-4d9a-acd6-6e2a0f1944b2.png)

nodet-opcua always returns an empty list arrayForInputArgumentDiagnosticInfo. 

![image](https://user-images.githubusercontent.com/680220/144423412-2429085b-9728-4b17-9993-9b6e76212a31.png)

hoancea:
I could assist and create a PR, however I would need to know which solution you would prefer, letting the consumer of the library check the `returnDiagnostics` flag or taking care of it in the final response generation.

erossignon:
Thank !
The second approach would be nice
>taking care of it in the final response generation

so that the diagnostic get filtered out before the response is sent if the client claimed did not request it in the `return diagnostics` field of the request header.
This will also not break the existing API. 

The other solution would require that `bindVariable` method signature changes

PR #1101, opened 2021-12-08
fix for #1093. filter out unrequested diagnostic information.

hoancea:
This PR corresponds to issue #1093. The diagnostic information is filtered out in the final response generation based on the bit values set in the `requestHeader` (`returnDiagnostics`). The translation is according to [this](https://reference.opcfoundation.org/v104/Core/docs/Part4/7.28/) specification.

### DIFF
--- a/packages/node-opcua-data-model/source/diagnostic_info.ts
+++ b/packages/node-opcua-data-model/source/diagnostic_info.ts
@@ -141,23 +141,15 @@ export class DiagnosticInfo extends BaseUAObject {
         decodeDebug_DiagnosticInfo(this, stream, options);
     }
 
-    public static filterForResponse(diagnostic: DiagnosticInfo, mask: number): DiagnosticInfo {
+    public static filterForResponse(diagnostic: DiagnosticInfo, mask: number, enumeration: DiagnosticInfo_Mask): DiagnosticInfo {
         const options: DiagnosticInfoOptions = {
-            symbolicId: (mask & DiagnosticInfo_ServiceLevelMask.SymbolicId || mask & DiagnosticInfo_OperationLevelMask.SymbolicId)
-                ? diagnostic.symbolicId
-                : undefined,
-            localizedText: (mask & DiagnosticInfo_ServiceLevelMask.LocalizedText || mask & DiagnosticInfo_OperationLevelMask.LocalizedText)
-                ? diagnostic.localizedText
-                : undefined,
-            additionalInfo: (mask & DiagnosticInfo_ServiceLevelMask.AdditionalInfo || mask & DiagnosticInfo_OperationLevelMask.AdditionalInfo)
-                ? diagnostic.additionalInfo
-                : undefined,
-            innerStatusCode: (mask & DiagnosticInfo_ServiceLevelMask.InnerStatusCode || mask & DiagnosticInfo_OperationLevelMask.InnerStatusCode)
-                ? diagnostic.innerStatusCode
-                : undefined,
-            innerDiagnosticInfo: (mask & DiagnosticInfo_ServiceLevelMask.InnerDiagnostics || mask & DiagnosticInfo_OperationLevelMask.InnerDiagnostics)
-                ? diagnostic.innerDiagnosticInfo
-                : (diagnostic.innerDiagnosticInfo ? DiagnosticInfo.filterForResponse(diagnostic.innerDiagnosticInfo, mask) : undefined),
+            symbolicId: (mask & enumeration.SymbolicId) ? diagnostic.symbolicId: undefined,
+            localizedText: (mask & enumeration.LocalizedText) ? diagnostic.localizedText: undefined,
+            additionalInfo: (mask & enumeration.AdditionalInfo) ? diagnostic.additionalInfo: undefined,
+            innerStatusCode: (mask & enumeration.InnerStatusCode) ? diagnostic.innerStatusCode: undefined,
+            innerDiagnosticInfo: (mask & enumeration.InnerDiagnostics) ? diagnostic.innerDiagnosticInfo : (
+                diagnostic.innerDiagnosticInfo ? DiagnosticInfo.filterForResponse(diagnostic.innerDiagnosticInfo, mask, enumeration) : undefined
+            ),
         }
         return new DiagnosticInfo(options);
     }
@@ -193,14 +185,17 @@ export enum DiagnosticInfo_OperationLevelMask {
     InnerDiagnostics = 0x0200
 }
 
+type DiagnosticInfo_Mask = typeof DiagnosticInfo_ServiceLevelMask | typeof DiagnosticInfo_OperationLevelMask;
+
 export const RESPONSE_DIAGNOSTICS_MASK_ALL = 0x3FF;
 
-export function filterDiagnosticInfoLevel(returnDiagnostics: number, diagnostic: DiagnosticInfo | null): DiagnosticInfo | null {
+export function filterDiagnosticInfoLevel(returnDiagnostics: number, diagnostic: DiagnosticInfo | null, level: "service" | "operation"): DiagnosticInfo | null {
     if (!diagnostic) {
         return null;
     }
 
-    return DiagnosticInfo.filterForResponse(diagnostic, returnDiagnostics);
+    const enumeration = level === "service" ? DiagnosticInfo_ServiceLevelMask : DiagnosticInfo_OperationLevelMask;
+    return DiagnosticInfo.filterForResponse(diagnostic, returnDiagnostics, enumeration);
 }
 
 export enum DiagnosticInfo_EncodingByte {
--- a/packages/node-opcua-data-model/source/diagnostic_info.ts
+++ b/packages/node-opcua-data-model/source/diagnostic_info.ts
@@ -141,25 +141,17 @@ export class DiagnosticInfo extends BaseUAObject {
         decodeDebug_DiagnosticInfo(this, stream, options);
     }
 
-    public filterForResponse(fields: number): void {
-        const schema = schemaDiagnosticInfo;
-        if (!(fields & DiagnosticInfo_ResponseDiagnostics.SymbolicId)) {
-            this.symbolicId = initialize_field(schema.fields[0], undefined) as Int32;
+    public static filterForResponse(diagnosticInfo: DiagnosticInfo, mask: number): DiagnosticInfo {
+        const options: DiagnosticInfoOptions = {
+            symbolicId: (mask & DiagnosticInfo_ResponseDiagnostics.SymbolicId) ? diagnosticInfo.symbolicId : undefined,
+            localizedText: (mask & DiagnosticInfo_ResponseDiagnostics.LocalizedText) ? diagnosticInfo.localizedText : undefined,
+            additionalInfo: (mask & DiagnosticInfo_ResponseDiagnostics.AdditionalInfo) ? diagnosticInfo.additionalInfo : undefined,
+            innerStatusCode: (mask & DiagnosticInfo_ResponseDiagnostics.InnerStatusCode) ? diagnosticInfo.innerStatusCode : undefined,
+            innerDiagnosticInfo: (mask & DiagnosticInfo_ResponseDiagnostics.InnerDiagnostics)
+                ? diagnosticInfo.innerDiagnosticInfo
+                : (diagnosticInfo.innerDiagnosticInfo ? DiagnosticInfo.filterForResponse(diagnosticInfo.innerDiagnosticInfo, mask) : undefined),
         }
-        if (!(fields & DiagnosticInfo_ResponseDiagnostics.LocalizedText)) {
-            this.localizedText = initialize_field(schema.fields[3], undefined) as Int32;
-        }
-        if (!(fields & DiagnosticInfo_ResponseDiagnostics.AdditionalInfo)) {
-            this.additionalInfo = initialize_field(schema.fields[4], undefined) as UAString;
-        }
-        if (!(fields & DiagnosticInfo_ResponseDiagnostics.InnerStatusCode)) {
-            this.innerStatusCode = initialize_field(schema.fields[5], undefined) as StatusCode;
-        }
-        if (!(fields & DiagnosticInfo_ResponseDiagnostics.InnerDiagnostics)) {
-            this.innerDiagnosticInfo = initialize_field(schema.fields[6], undefined) as DiagnosticInfo;
-        } else if (this.innerDiagnosticInfo) {
-            this.innerDiagnosticInfo.filterForResponse(fields);
-        }
+        return new DiagnosticInfo(options);
     }
 }
 
@@ -190,17 +182,26 @@ export enum DiagnosticInfo_ResponseDiagnostics {
     InnerDiagnostics = 0x10
 }
 
+// in the case of operation diagnostics being requested, we need to determine how many bits we need to shift to the right
+// we do this in order to prevent code duplication which would've been introduced by creating another enum
+const enumNumberValues = Object.values(DiagnosticInfo_ResponseDiagnostics).filter((value) => typeof value === "number" && value !== 0);
+// the number of bits we need to shift is calculated by counting the non-zero elements in the enumeration
+export const OPERATION_DIAGNOSTICS_BITS_TO_SHIFT: number = enumNumberValues.length;
+
 export const RESPONSE_DIAGNOSTICS_MASK_ALL =
     DiagnosticInfo_ResponseDiagnostics.SymbolicId | DiagnosticInfo_ResponseDiagnostics.LocalizedText | DiagnosticInfo_ResponseDiagnostics.AdditionalInfo |
     DiagnosticInfo_ResponseDiagnostics.InnerStatusCode | DiagnosticInfo_ResponseDiagnostics.InnerDiagnostics;
 
 export function filterDiagnosticInfoLevel(returnDiagnostics: number, diagnosticInfo: DiagnosticInfo | (DiagnosticInfo | null)[] | null): void {
+    if (!diagnosticInfo) {
+        return;
+    }
     const items = Array.isArray(diagnosticInfo) ? diagnosticInfo : [diagnosticInfo];
-    for (const entry of items) {
+    for (let entry of items) {
         if (!entry) {
             continue;
         }
-        entry.filterForResponse(returnDiagnostics);
+        entry = DiagnosticInfo.filterForResponse(entry, returnDiagnostics);
     }
 }
 
--- a/packages/node-opcua-data-model/source/diagnostic_info.ts
+++ b/packages/node-opcua-data-model/source/diagnostic_info.ts
@@ -140,6 +140,27 @@ export class DiagnosticInfo extends BaseUAObject {
     public decodeDebug(stream: BinaryStream, options: DecodeDebugOptions): void {
         decodeDebug_DiagnosticInfo(this, stream, options);
     }
+
+    public filterForResponse(fields: number): void {
+        const schema = schemaDiagnosticInfo;
+        if (!(fields & DiagnosticInfo_ResponseDiagnostics.SymbolicId)) {
+            this.symbolicId = initialize_field(schema.fields[0], undefined) as Int32;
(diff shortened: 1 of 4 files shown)


packages/node-opcua-server/source/opcua_server.ts
```diff
@@ -3339,13 +3341,42 @@ export class OPCUAServer extends OPCUABaseServer {
...
                     }
                 );
             }
         );
     }
 
+    private _filterDiagnosticInfo(returnDiagnostics: number, response: CallResponse): void {
+        if ((RESPONSE_DIAGNOSTICS_MASK_ALL * DiagnosticInfo_ResponseDiagnosticsLevel.Service & returnDiagnostics) && response.responseHeader.serviceDiagnostics) {
+            filterDiagnosticInfoLevel(
+                returnDiagnostics / DiagnosticInfo_ResponseDiagnosticsLevel.Service, response.responseHeader.serviceDiagnostics
```
erossignon: I would use shift operator as in javascriopt returnDiagnostics / DiagnosticInfo_ResponseDiagnosticsLevel.Service  will be turned to a float number with decimal if a low bit exists

packages/node-opcua-data-model/source/diagnostic_info.ts
```diff
@@ -140,6 +140,27 @@ export class DiagnosticInfo extends BaseUAObject {
     public decodeDebug(stream: BinaryStream, options: DecodeDebugOptions): void {
         decodeDebug_DiagnosticInfo(this, stream, options);
     }
+
+    public filterForResponse(fields: number): void {
```
erossignon: we could turn this to  static returns that return a new DiagnosticInfo

````javascript
static filterForResponse(diagnosticInfo: DiagnosticInfo, mask: number): DiagnosticInfo{
    const options = {
         symbolicId: (mask & (DiagnosticInfoOperationLevelMask.SymbolicId | DiagnosticInfoServiceLevelMask.SymbolicId ) : diagnosticInfo.symbolicId | undefined,
         localizedText : (mask & (DiagnosticInfoOperationLevelMask.LocalizedText | DiagnosticInfoServiceLevelMask.LocalizedText) : diagnosticInfo.localizedText | undefined,
... etc...
    }
    return new DiagnosticInfo(diagnosticInfo);
}
```` 

this will avoid using the initialize_field with schema.fields[0] and risk of error choosing the right index.

File: packages/node-opcua-data-model/source/diagnostic_info.ts [hunk 3]
```diff
@@ -155,6 +176,34 @@ export interface DiagnosticInfoOptions {
     innerDiagnosticInfo?: DiagnosticInfo;
 }
 
+export enum DiagnosticInfo_ResponseDiagnosticsLevel {
```
erossignon: According to the specs the  bit value are :

https://reference.opcfoundation.org/v104/Core/docs/Part4/7.28/
```` 
0x0000 0001 ServiceLevel / SymbolicId
   0x0000 0002 ServiceLevel / LocalizedText
   0x0000 0004 ServiceLevel / AdditionalInfo
   0x0000 0008 ServiceLevel / Inner StatusCode
   0x0000 0010 ServiceLevel / Inner Diagnostics
   0x0000 0020 OperationLevel / SymbolicId
   0x0000 0040 OperationLevel / LocalizedText
   0x0000 0080 OperationLevel / AdditionalInfo
   0x0000 0100 OperationLevel / Inner StatusCode
   0x0000 0200 OperationLevel / Inner Diagnostics
````

why not use 
```` 
export enum DiagnosticInfoServiceLevelMask {
    None = 0x00,
    SymbolicId = 0x01,
    LocalizedText = 0x02,
    AdditionalInfo = 0x04,
    InnerStatusCode = 0x08,
    InnerDiagnostics = 0x10
}
export enum DiagnosticInfoOperationLevelMask {
    SymbolicId = 0x020,
    LocalizedText = 0x040,
    AdditionalInfo = 0x080,
    InnerStatusCode = 0x0100,
    InnerDiagnostics = 0x0200
}

File: packages/node-opcua-server/source/opcua_server.ts [hunk 4]
```diff
@@ -3339,13 +3341,42 @@ export class OPCUAServer extends OPCUABaseServer {
...
+            );
+        } else {
+            response.responseHeader.serviceDiagnostics = null;
+        }
+
+        if (RESPONSE_DIAGNOSTICS_MASK_ALL * DiagnosticInfo_ResponseDiagnosticsLevel.Operation & returnDiagnostics) {
+            if (response.diagnosticInfos && response.diagnosticInfos.length > 0) {
+                filterDiagnosticInfoLevel(returnDiagnostics / DiagnosticInfo_ResponseDiagnosticsLevel.Operation, response.diagnosticInfos);
```
erossignon: `returnDiagnostics / DiagnosticInfo_ResponseDiagnosticsLevel.Operation` is a rational number, we would need a integer

erossignon (changes requested):

packages/node-opcua-data-model/source/diagnostic_info.ts
hoancea: I have implemented the enum this way in order to prevent code duplication

packages/node-opcua-server/source/opcua_server.ts
hoancea: solved in commit `de8d999ce818fe2069604d1c19bf8805f0cf7d3a`

packages/node-opcua-data-model/source/diagnostic_info.ts
hoancea: solved in commit `de8d999ce818fe2069604d1c19bf8805f0cf7d3a`

packages/node-opcua-server/source/opcua_server.ts
hoancea: solved in commit `de8d999ce818fe2069604d1c19bf8805f0cf7d3a`

packages/node-opcua-data-model/source/diagnostic_info.ts
erossignon: It would reflect more the specification.

packages/node-opcua-data-model/source/diagnostic_info.ts
```diff
@@ -155,6 +168,43 @@ export interface DiagnosticInfoOptions {
...
+    AdditionalInfo = 0x04,
+    InnerStatusCode = 0x08,
+    InnerDiagnostics = 0x10
+}
+
+// in the case of operation diagnostics being requested, we need to determine how many bits we need to shift to the right
+// we do this in order to prevent code duplication which would've been introduced by creating another enum
+const enumNumberValues = Object.values(DiagnosticInfo_ResponseDiagnostics).filter((value) => typeof value === "number" && value !== 0);
+// the number of bits we need to shift is calculated by counting the non-zero elements in the enumeration
+export const OPERATION_DIAGNOSTICS_BITS_TO_SHIFT: number = enumNumberValues.length;
```
erossignon: The needed value can be hardcoded I think ( we can count to 5) ... A unit test should detect if we made it wrong.

packages/node-opcua-data-model/source/diagnostic_info.ts
```diff
@@ -155,6 +168,43 @@ export interface DiagnosticInfoOptions {
...
+
+export const RESPONSE_DIAGNOSTICS_MASK_ALL =
```
erossignon: 0x1F will do

packages/node-opcua-data-model/source/diagnostic_info.ts
```diff
@@ -155,6 +168,43 @@ export interface DiagnosticInfoOptions {
...
+export function filterDiagnosticInfoLevel(returnDiagnostics: number, diagnosticInfo: DiagnosticInfo | (DiagnosticInfo | null)[] | null): void {
+    if (!diagnosticInfo) {
+        return;
+    }
+    const items = Array.isArray(diagnosticInfo) ? diagnosticInfo : [diagnosticInfo];
+    for (let entry of items) {
+        if (!entry) {
+            continue;
+        }
+        entry = DiagnosticInfo.filterForResponse(entry, returnDiagnostics);
```
erossignon: This has no effect on the diagnostic info if this is a array... I guess one unit test is missing for this case.

packages/node-opcua-data-model/test/test_diagnostic_info.js
```diff
@@ -121,6 +121,77 @@ describe("DiagnosticInfo", function () {
...
 
+    it("should not strip away diagnostic information if requested", function () {
+        let diag = new DiagnosticInfo({
+            localizedText: 2345,
+            symbolicId: 3456,
+            additionalInfo: "test",
+            innerStatusCode: StatusCodes.Bad,
+            innerDiagnosticInfo: new DiagnosticInfo({ additionalInfo: "test 2" })
+        });
+        const fields = DiagnosticInfo_ResponseDiagnostics.SymbolicId
```
erossignon: We could rename `fields` to `serviceLevelMask` to better reflect what the value is

packages/node-opcua-data-model/test/test_diagnostic_info.js
```diff
@@ -121,6 +121,77 @@ describe("DiagnosticInfo", function () {
...
+        let diag = new DiagnosticInfo({
+            localizedText: 2345,
+            symbolicId: 3456,
+            additionalInfo: "test",
+            innerStatusCode: StatusCodes.Bad,
+            innerDiagnosticInfo: new DiagnosticInfo({ additionalInfo: "test 2" })
+        });
+
+        diag = DiagnosticInfo.filterForResponse(diag, DiagnosticInfo_ResponseDiagnostics.None);
+        diag.localizedText.should.equal(-1);
```
erossignon, line 160: I am expecting diag.localizredText to be undefined .  I was expecting 
````
should.not.exist(diag.localizedTex);
```` 
?
( keep it this way for the moment, there might be a unforeseen bug in the DiagnosticInfo constructor that you've highlighted here )

File: packages/node-opcua-data-model/test/test_diagnostic_info.js [hunk 10]
```diff
@@ -121,6 +121,77 @@ describe("DiagnosticInfo", function () {
...
+
+    it("should strip away unrequested details", function () {
+        let diag = new DiagnosticInfo({
+            localizedText: 2345,
+            symbolicId: 3456,
+            additionalInfo: "test",
+            innerStatusCode: StatusCodes.Bad,
+            innerDiagnosticInfo: new DiagnosticInfo({ additionalInfo: "test 2", innerStatusCode: StatusCodes.Bad, symbolicId: 34567 })
+        });
+        const fields = DiagnosticInfo_ResponseDiagnostics.LocalizedText
```
erossignon: serviceLevelMask

packages/node-opcua-data-model/test/test_diagnostic_info.js
```diff
@@ -121,6 +121,77 @@ describe("DiagnosticInfo", function () {
...
+        diag.additionalInfo.should.equal("test");
+        diag.innerStatusCode.should.equal(StatusCodes.Good); // 'StatusCodes.Good' is the default value for 'innerStatusCode'
+
+        diag.innerDiagnosticInfo.localizedText.should.equal(-1);
+        diag.innerDiagnosticInfo.symbolicId.should.equal(34567);
+        should(diag.innerDiagnosticInfo.additionalInfo).equal("test 2");
+        diag.innerDiagnosticInfo.innerStatusCode.should.equal(StatusCodes.Good); // 'StatusCodes.Good' is the default value for 'innerStatusCode'
+        should(diag.innerDiagnosticInfo.innerDiagnosticInfo).equal(null);
+    });
+
```
erossignon, line 196: Please add unit test for filterDiagnosticInfoLevel as well

packages/node-opcua-server/source/opcua_server.ts
```diff
@@ -3339,13 +3342,40 @@ export class OPCUAServer extends OPCUABaseServer {
...
                         });
+                        this._filterDiagnosticInfo(request.requestHeader.returnDiagnostics, response);
                         sendResponse(response);
                     }
                 );
             }
         );
     }
 
+    private _filterDiagnosticInfo(returnDiagnostics: number, response: CallResponse): void {
```
erossignon: I would make this method an external function. ( `this` is not used )

erossignon: You can probably make filterDiagnosticInfoLevel works with Service AND operation level  internally , so we do not have to put the complexity to shift the bits at this higher level.

packages/node-opcua-server/source/opcua_server.ts
```diff
@@ -3339,13 +3342,40 @@ export class OPCUAServer extends OPCUABaseServer {
...
+                    if (entry.inputArgumentDiagnosticInfos) {
+                        filterDiagnosticInfoLevel(returnDiagnostics >> OPERATION_DIAGNOSTICS_BITS_TO_SHIFT, entry.inputArgumentDiagnosticInfos);
+                    } else {
+                        entry.inputArgumentDiagnosticInfos = [];
+                    }
+                }
+            }
+        }
+    }
+
```
erossignon: unit test ?

erossignon (changes requested):

packages/node-opcua-data-model/test/test_diagnostic_info.js
erossignon: image

This is more complex than I though, the localizedText of the DiagnosticInfo is a index pointing to a string in the string table of the response Header

image

erossignon (changes requested):
I also see a merge commit with master ... 
Would you be able to rewrite the commits to keep a linear history? This would mean using git rebase instead of git merge when resyncing with main branch.

packages/node-opcua-server/source/opcua_server.ts
```diff
@@ -677,6 +679,34 @@ function validate_security_endpoint(
     return { errCode: StatusCodes.Good, endpoint: endpoints_matching_security_policy[0] };
 }
 
+export function filterDiagnosticInfo(returnDiagnostics: number, response: CallResponse): void {
+    if (RESPONSE_DIAGNOSTICS_MASK_ALL & returnDiagnostics) {
+        response.responseHeader.serviceDiagnostics = filterDiagnosticInfoLevel(returnDiagnostics, response.responseHeader.serviceDiagnostics, "service");
+
+        if (response.diagnosticInfos && response.diagnosticInfos.length > 0) {
+            response.diagnosticInfos.forEach(
```
erossignon: I would use
 ```typescript
response.diagnosticInfos = response.diagnosticInfos.map(
        (d)=>filterDiagnosticOperationLevel(returnDiagnostics,d));

```` 

by creating a small helper  filterDiagnosticOperationLevel
and using map instead of forEach.

File: packages/node-opcua-server/source/opcua_server.ts [hunk 15]
```diff
@@ -677,6 +679,34 @@ function validate_security_endpoint(
...
+                    array[index] = filterDiagnosticInfoLevel(returnDiagnostics, diagnostic, "operation")
+            );
+        } else {
+            response.diagnosticInfos = [];
+        }
+
+        if (response.results) {
+            for (const entry of response.results) {
+                if (entry.inputArgumentDiagnosticInfos && entry.inputArgumentDiagnosticInfos.length > 0) {
+                    entry.inputArgumentDiagnosticInfos.forEach(
```
erossignon: same here : 

 ```typescript
response.diagnosticInfos = response.diagnosticInfos.map(
        (d)=>filterDiagnosticOperationLevel(returnDiagnostics,d));

```` 

by creating a small helper  filterDiagnosticOperationLevel
and using map instead of forEach.

File: packages/node-opcua-server/test/test_server_diagnostic_filter.js [hunk 16]
```diff
@@ -0,0 +1,65 @@
...
+            },
+            results: [{ inputArgumentDiagnosticInfos: [
+                new DiagnosticInfo({ additionalInfo: "input argument", innerStatusCode: StatusCodes.Bad, symbolicId: 34567 })
+            ] }],
+            diagnosticInfos: [new DiagnosticInfo({ additionalInfo: "diagnostic infos", innerStatusCode: StatusCodes.Bad, symbolicId: 34567 })]
+        });
+    });
+
+    it("should filter diagnostic information", function () {
+        const serviceLevelMask = DiagnosticInfo_ServiceLevelMask.SymbolicId
```
erossignon: THis is more than just a serviceLevelMask  as it contains bits from the OperationLevelMast, Can we rename it as `levelMask`

packages/node-opcua-data-model/source/diagnostic_info.ts
```diff
@@ -140,6 +140,19 @@ export class DiagnosticInfo extends BaseUAObject {
     public decodeDebug(stream: BinaryStream, options: DecodeDebugOptions): void {
         decodeDebug_DiagnosticInfo(this, stream, options);
     }
+
+    public static filterForResponse(diagnostic: DiagnosticInfo, mask: number, enumeration: DiagnosticInfo_Mask): DiagnosticInfo {
```
erossignon: I would rename `enumeration` to `diagnosticInfoMask`
and `mask` to `requestedDiagnostics`.

packages/node-opcua-data-model/source/diagnostic_info.ts
```diff
@@ -155,6 +168,36 @@ export interface DiagnosticInfoOptions {
...
+    AdditionalInfo = 0x080,
+    InnerStatusCode = 0x0100,
+    InnerDiagnostics = 0x0200
+}
+
+type DiagnosticInfo_Mask = typeof DiagnosticInfo_ServiceLevelMask | typeof DiagnosticInfo_OperationLevelMask;
+
+export const RESPONSE_DIAGNOSTICS_MASK_ALL = 0x3FF;
+
+export function filterDiagnosticInfoLevel(returnDiagnostics: number, diagnostic: DiagnosticInfo | null, level: "service" | "operation"): DiagnosticInfo | null {
```
erossignon: I would replace:

```` 
level: "service" | "operation"
```` 
with
```` 
diagnosticInfoMask: DiagnosticInfo_Mask
```` 

so that
```` 
 const enumeration = level === "service" ? DiagnosticInfo_ServiceLevelMask : DiagnosticInfo_OperationLevelMask;
    return DiagnosticInfo.filterForResponse(diagnostic, returnDiagnostics, enumeration);
```` 
becomes
````
    return DiagnosticInfo.filterForResponse(diagnostic, returnDiagnostics, diagnosticInfoMask);
````

File: packages/node-opcua-server/source/opcua_server.ts [hunk 19]
```diff
@@ -677,6 +679,34 @@ function validate_security_endpoint(
     return { errCode: StatusCodes.Good, endpoint: endpoints_matching_security_policy[0] };
 }
 
+export function filterDiagnosticInfo(returnDiagnostics: number, response: CallResponse): void {
+    if (RESPONSE_DIAGNOSTICS_MASK_ALL & returnDiagnostics) {
+        response.responseHeader.serviceDiagnostics = filterDiagnosticInfoLevel(returnDiagnostics, response.responseHeader.serviceDiagnostics, "service");
```
erossignon: replace "service" with `DiagnosticInfo_ServiceLevelMask `

packages/node-opcua-server/source/opcua_server.ts
```diff
@@ -677,6 +679,34 @@ function validate_security_endpoint(
...
+
+        if (response.diagnosticInfos && response.diagnosticInfos.length > 0) {
+            response.diagnosticInfos.forEach(
+                (diagnostic: DiagnosticInfo | null, index: number, array: (DiagnosticInfo | null)[]) =>
+                    array[index] = filterDiagnosticInfoLevel(returnDiagnostics, diagnostic, "operation")
```
erossignon: replace "operation" with replace "service" with DiagnosticInfo_OperationLevelMask

erossignon (changes requested):
Thank you Bodgan for taking care of fixing my previous comments.
We are almost here, I just spotted a few code simplifications that we could still do.

PR #1121, opened 2022-03-06
Improve load node set

erossignon:


erossignon:
Error seems to be specific to the sample demo code . I have put my comment in the PR
https://github.com/AndreasHeine/SampleServer-node-opcua/pull/123

PR #1123, opened 2022-03-15
improve message of OffNormalAlarm

AndreasHeine:


AndreasHeine:
currently:
image
new:
image

probably the enum-string with that index would be more user-friendly...

tell my what you think!

erossignon:
change has been superceeded with recent work in [1e849f674]

Issue #1136, opened 2022-04-11
node-opcua-server crashes if invalid user identity token certificate is given


SanderDeWaal1992:
**Current behavior**

The server crashes when I connect with a client with an invalid user identity token certificate.

**Describe the bug**

The server crashes when I connect with a client with an invalid user token certificate. It is the assertion here:
https://github.com/node-opcua/node-opcua/blob/bb9e5c1160a4362fc33a124ae1476d6bbf102157/packages/node-opcua-server/source/opcua_server.ts#L2079

As you can see there, the error is not used. It immediately crashes because it assumes that the certificate is valid.

**Step by step instruction to reproduce**

1. Connect to the server with an invalid user token certificate (I encountered it while using the node-opcua-client with an PEM formatted certificate as the userIdentity certificate instead of a der formatted certificate, but I believe it can basically be any invalid certificate)
2. See the server crashing with a non-informative 'expecting statusCode' error

**Expected behavior**

Clients can not be trusted so the server should not crash. The error should be logged.
This to prevent easy DOS attacks and guaranteeing that the correct operation of the rest of the system can not be harmed by one client accidentally having an invalid certificate (and not even having a clue what client it is due to lack of logging of the error).  

- **node-opcua version:** :
encountered it on 2.59.0 of node-opcua-server and it is also on current master

- **Node:**
16.13.1

SanderDeWaal1992:
I now also encountered a comparable problem when supplying an invalid subjectAltName extension. In that case the server also crashes because it assumes the certificate contains the expected fields without checking it. 

Does node-opcua have the intention to prevent these kind of easy DOS attacks or protection against accidental misconfigurations (by users of our product over which we have no control)?

SanderDeWaal1992:
I will start working on creating a PR to solve above issues.

erossignon:
fixed in node-opcua@2.70.0

PR #1142, opened 2022-05-04
fix for #1136. fix server crashing on invalid certificate

hoancea:
fix for #1136. See steps listed on the issue in order to replicate the issue


hoancea:
I signed the CLA twice already, yet it still appears as not signed on the issue, not sure why.

packages/node-opcua-server/source/opcua_server.ts
```diff
@@ -1502,7 +1503,11 @@ export class OPCUAServer extends OPCUABaseServer {
         assert(userTokenSignature.signature instanceof Buffer, "expecting userTokenSignature to be a Buffer");
 
         // verify proof of possession by checking certificate signature & server nonce correctness
-        if (!verifySignature(serverCertificate, nonce, userTokenSignature, certificate, securityPolicy)) {
+        try {
```
erossignon: Can we fix verifySignature instance to make it return false instead of throwing ?

erossignon (changes requested):

packages/node-opcua-server/source/opcua_server.ts
hoancea: solved

packages/node-opcua-secure-channel/source/security_policy.ts
```diff
@@ -31,11 +31,15 @@ import {
...
 } from "node-opcua-crypto";
 import { EncryptBufferFunc, SignBufferFunc } from "node-opcua-chunkmanager";
+import { make_warningLog } from "node-opcua-debug";
 
 // tslint:disable:no-empty
 function errorLog(...args: any[]) {
     /** */
 }
+const warningLog = make_warningLog(__filename);
+const doTraceChunk = process.env.NODEOPCUADEBUG && process.env.NODEOPCUADEBUG.indexOf("CHUNK") >= 0;
```
erossignon: CHUNK is used for tracing low level packet chunking on the transport layer

packages/node-opcua-secure-channel/source/security_policy.ts
```diff
@@ -584,7 +588,14 @@ export function verifySignature(
     assert(signature.signature instanceof Buffer);
     // This parameter is calculated by appending the clientNonce to the clientCertificate
     const dataToVerify = Buffer.concat([chain[0], receiverNonce]);
-    return cryptoFactory.asymmetricVerify(dataToVerify, signature.signature, senderCertificate);
+    try {
+        return cryptoFactory.asymmetricVerify(dataToVerify, signature.signature, senderCertificate);
+    } catch (e) {
+        if (doTraceChunk) {
```
erossignon: Let's remove this if statemtn, a warning can be emitted regardless.

erossignon (changes requested):

packages/node-opcua-secure-channel/source/security_policy.ts
erossignon: I would remove the use of doTraceChunk  alltogether

packages/node-opcua-secure-channel/source/security_policy.ts
hoancea: fixed in commit d413004

packages/node-opcua-secure-channel/source/security_policy.ts
hoancea: fixed in commit d413004

PR #1179, opened 2022-08-07
Add a max retry to findServers and findServersOnNetwork to prevent infinite retries

daveycodes:
A max retry should be used so that the connect attempt will throw an error if the discovery server cannot be reached. Without a max retry the callback is never executed, leaving any callers of these functions without any way to handle this type of scenario.


erossignon:
Thank you

PR #1184, opened 2022-08-13
Fix eventFilter selectClause in alarm tools

### LINKED ISSUE #1183: Alarm_Tools EventFilter SelectClause contains two entries for ConditionId, instead of one
**Current behavior**
In ```client_alarm_tools_acknowledge_all_conditions``` and ```client_alarm_tools``` the SelectClause in the created EventFilter contains two entries for the ConditionId instead of one. A wrong one and a correct one. I found out, because I used the same pattern "extractConditionFields" -> "constructEventFilter" in the node-red opc ua library. I am not sure to what issues that leads in those two components.

**Describe the bug**
In ```client_alarm_tools_acknowledge_all_conditions``` and ```client_alarm_tools``` the SelectClause in the created EventFilter contains two entries for the ConditionId insetad of one. A wrong one and a correct one. I found out, because I used the same pattern "extractConditionFields" -> "constructEventFilter" in the node-red opc ua library. I am not sure to what issues that leads in those two components.

**Step by step instruction to reproduce**

- Execute "findActiveConditions" and inspect eventFilter
- Execute "installAlarmMonitoring" and inspect eventFilter

**Expected behavior**

**Context**

- (x ) my request is related to node-opcua acting as an OPCUA CLIENT
- ( ) my request is related to node-opcua acting as an OPCUA SERVER

- ( ) I have installed node-opcua from source ( using git clone)
- (x) I have installed node-opcua as a package ( using npm install )
- ( ) I am using an application that uses node-opcua

    - ( ) node-red
    - ( ) other : please specify

Koseng:
This should fix #1183. 

I did not test the detailed behavior before and after.

### DIFF
--- a/packages/node-opcua-client/source/alarms_and_conditions/client_alarm.ts
+++ b/packages/node-opcua-client/source/alarms_and_conditions/client_alarm.ts
@@ -114,6 +114,5 @@ export function fieldsToJson(fields: string[], eventFields: Variant[]): EventStu
         const variant = eventFields[index];
         setProperty(data, fields[index], variant);
     }
-    setProperty(data, "conditionId", eventFields[eventFields.length - 1]);
     return data;
 }
--- a/packages/node-opcua-client/source/alarms_and_conditions/client_alarm_tools.ts
+++ b/packages/node-opcua-client/source/alarms_and_conditions/client_alarm_tools.ts
@@ -84,7 +84,8 @@ export async function installAlarmMonitoring(session: ClientSession): Promise<Cl
 
     const fields = await extractConditionFields(session, "AlarmConditionType");
 
-    const eventFilter = constructEventFilter(fields, [resolveNodeId("AcknowledgeableConditionType")]);
+    // Remove ConditionId in fields. SelectClause entry for ConditionId is added via conditionTypes argument.
+    const eventFilter = constructEventFilter(fields.slice(0, -1), [resolveNodeId("AcknowledgeableConditionType")]);
 
     const monitoringParameters: MonitoringParametersOptions = {
         discardOldest: false,
--- a/packages/node-opcua-client/source/alarms_and_conditions/client_alarm_tools_acknowledge_all_conditions.ts
+++ b/packages/node-opcua-client/source/alarms_and_conditions/client_alarm_tools_acknowledge_all_conditions.ts
@@ -65,7 +65,8 @@ export async function findActiveConditions(session: ClientSession): Promise<Even
 
     const fields = await extractConditionFields(session, "AcknowledgeableConditionType");
 
-    const eventFilter = constructEventFilter(fields, [resolveNodeId("AcknowledgeableConditionType")]);
+    // Remove ConditionId in fields. SelectClause entry for ConditionId is added via conditionTypes argument.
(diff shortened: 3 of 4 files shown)


erossignon:
[Koseng](https://github.com/Koseng), please add a unit test that demonstrates the issue.

Koseng:
I had now time and a closer look. It was working so far despite the bug, because later on a workaround was implemented.

This is the issue. For the ConditionId, two select clauses are created. So there is one selectClause more than number of fields.

|Index | Fields            | SelectClauses| 
|-------|------|-----------|
| 0      | A      | Correct SelectClause for A| 
| 1      | B      | Correct SelectClause for B | 
| 2      | C      | Correct SelectClause for C| 
| 3      | D      | Correct SelectClause for D | 
| 4      | ConditionID      | Wrong SelectClause for CondId | 
| 5      |          | Correct SelectClause for CondId | 

The screenshot shows the code with current active workaround.
With this PR only the yellow part in ```fieldsToJson``` is necessary. The green marked one can be removed.

The green marked one was needed additionally to workaround the bug. It copies the result of the last (correct) SelectClause to the ConditionId field. 

image

I will add an e2e Test and also remove the green marked line.

erossignon:
[Koseng](https://github.com/Koseng) let me propose a slightly different fix that addresses the root cause of the issue you have identified.
we will make the "ConditionId" an explicitly field and provide an appropriate conversion to SimpleAttributeOperand when converted to selectClause.

Koseng:
@erossignon, ok I had a look. This solution is also fine, you just detect "ConditionId" in the fields list as a special case.

Another topic is how to create the whereClause in "constructEventFilter". I think the whereClause could/should be auto created considering a given type and automatically it's subtypes. Like that it is implemented in other libraries as well.

"I want to have all events of ConditionType" -> This should include the events from types which inherit from ConditionType.

I implemented it for the node-red opc ua contrib library in, and planned to propose it for constructEventFilter as well:
https://github.com/mikakaraila/node-red-contrib-opcua/pull/468

Have a look at: ```async function create_where_clause_from_eventtype(eventTypeId)```

Finally it would be great to have "conditionTypes" renamed to more like "eventTypes". To make clear the helper method constructEventFilter can be used for all kind of events not only condition related ones.

erossignon:
The whereClause is using the  OfType(MyEventType) operator that do exactly whatyou are looking for , without the need to use the InList([A,B,C,]), and explicitly enumerate all possibles Subtypes;


It's still work in progress on my side and I am planning to add a more sophisticated mecanims where we will be able to do
something like

Or(OfType(A),OfType(B)) or Or(OfType(A),OfType(B),InList([C,D,E,F]) etc ...
 

What you are doing by crafting the where clause directly on your side is fine.

erossignon:
> Finally it would be great to have "conditionTypes" renamed to more like "eventTypes". To make clear the helper method constructEventFilter can be used for all kind of events not only condition related ones.

I'll take your point and adjust my PR accordingly.

erossignon:
node-opcua@2.75.0 provides a bettter createEventFilter ,
It also permits complex WhereClauses to be easily crafted.

erossignon:
@Koseng , thank you for your support

PR #1187, opened 2022-08-26
fix: add missing chalk dependency

wolasss:
In multiple internal packages `chalk` is a missing dependency, causing the production build to fail: 

```
Error: Cannot find module 'chalk'
Require stack:
- /home/node/core/node_modules/node-opcua-service-discovery/dist/bonjour.js
- /home/node/core/node_modules/node-opcua-service-discovery/dist/index.js
- /home/node/core/node_modules/node-opcua/dist/index.js
- /home/node/core/index.js
```

I added missing dependency in the packages that use `chalk` but have no `chalk` dependency specified in `package.json`

### DIFF
--- a/packages/node-opcua-convert-nodeset-to-javascript/package.json
+++ b/packages/node-opcua-convert-nodeset-to-javascript/package.json
@@ -12,6 +12,7 @@
         "test-generated": "tsc --build tsconfig-generated.json"
     },
     "dependencies": {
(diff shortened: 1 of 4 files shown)


erossignon:
Quick Question, 
Most of the time this missing package is not a problem as we usually reference node-opcua or node-opcua-client which already include chalk.

How did you figure out that the chalk package was missing?

PR #1188, opened 2022-08-27
Fix producturi in session

### LINKED ISSUE #1170: OPC UA Non-Compliance: Endpoint description information returned by node-opcua server does not match.
NODE-OPCUA stack version : 2.72.2
NODEJS version : v16.15.1

Current Behavior:
 
Endpoint description information returned by node-opcua server in the GetEndPoint Response does not match with the Endpoint description information returned in the Create session response.
Snapshot of wireshark capture GetEndPoint response where the Product URI is returned
image

However in the below CreateSession response wireshark capture snapshot Product URI is empty

image

**Expected behavior**
As per the OPC UA compliance spec, endpoint description returned in the GetEndpointResponse and in the CreateSessionResponse by OPC UA server should exactly match .

Test case:
<html>
<body>

Err-023 | CTT | CreateSession | productUris contained in the returned SoftwareCertificates do not match the productUris in the EndpointDescription used by the Client to connect to the Server. |   | Client closes session calling CloseSession.Client logs and/or display an error to the end user.
-- | -- | -- | -- | -- | --

</body>
</html>

- ( ) my request is related to node-opcua acting as an OPCUA SERVER

JeanArhancet:
Hi,

This is my first MR for this project :) 

Fix #1170

### DIFF
--- a/packages/node-opcua-end2end-test/test/end_to_end/test_e2e_1170.ts
+++ b/packages/node-opcua-end2end-test/test/end_to_end/test_e2e_1170.ts
@@ -1,0 +1,79 @@
+const async = require("async");
+const should = require("should");
+const opcua = require("node-opcua");
+const OPCUAClient = opcua.OPCUAClient;
+const UserTokenPolicy = opcua.UserTokenPolicy;
+const { perform_operation_on_client_session } = require("../../test_helpers/perform_operation_on_client_session");
+// require("node-opcua-service-session").UserNameIdentityToken;
+
+const describe = require("node-opcua-leak-detector").describeWithLeakDetector;
+describe("Testing bug #1170", function() {
+
+    const port = 1170;
+    let server = null;
+    let endpointUrl = "";
+
+
+    before(async () => {
+
+
+        server = new opcua.OPCUAServer({
+            port
+        });
+
+        await server.start();
+
+        let endpoints = server._get_endpoints();
+        endpointUrl = endpoints[0].endpointUrl;
+
+
+    });
+    after(function(done) {
+        server.shutdown(done);
+    });
+
+    it("should productURI have the same value", (done) => {
+
+        // user1/password1
+
+        const client = opcua.OPCUAClient.create({
+            endpointMustExist: false,
+            requestedSessionTimeout: 60000
+        });
+
+        let the_session;
+        async.series([
+
+            function client_connect(callback) {
+                client.connect(endpointUrl, function(err) {
+                    if (err) {
+                        console.log(" cannot connect to endpoint :", endpointUrl);
+                    } else {
+                        console.log("connected !");
+                    }
+                    callback(err);
+                });
+      
+            },
+
+            function client_create_session(callback) {
+                
+                client.createSession({}, function(err, session) {
+                    if (!err) {
+                        the_session = session;
+                        should(client.endpoint.server.productUri).eql("NodeOPCUA-Server");
+                    }
+                    callback(err);
+                });
+            },
+
+            function(callback) {
+                the_session.close(callback);
+            },
+            function(callback) {
+                client.disconnect(callback);
+            }
+        ], done);
+    });
+});
+
--- a/packages/node-opcua-end2end-test/test/end_to_end/test_e2e_1170.ts
+++ b/packages/node-opcua-end2end-test/test/end_to_end/test_e2e_1170.ts
@@ -4,76 +4,33 @@ const opcua = require("node-opcua");
 const OPCUAClient = opcua.OPCUAClient;
 const UserTokenPolicy = opcua.UserTokenPolicy;
 const { perform_operation_on_client_session } = require("../../test_helpers/perform_operation_on_client_session");
-// require("node-opcua-service-session").UserNameIdentityToken;
 
 const describe = require("node-opcua-leak-detector").describeWithLeakDetector;
-describe("Testing bug #1170", function() {
-
+describe("Testing bug #1170", function () {
     const port = 1170;
-    let server = null;
-    let endpointUrl = "";
-
-
+    let server;
+    let endpointUrl;
     before(async () => {
-
-
         server = new opcua.OPCUAServer({
-            port
+            port,
+            serverInfo: {
+                productUri: "Mini NodeOPCUA-Server"
+            }
         });
 
         await server.start();
 
-        let endpoints = server._get_endpoints();
-        endpointUrl = endpoints[0].endpointUrl;
-
-
+        endpointUrl = server.getEndpointUrl();
     });
-    after(function(done) {
-        server.shutdown(done);
+    after(async () => {
+        await server.shutdown();
     });
 
-    it("should productURI have the same value", (done) => {
-
-        // user1/password1
-
-        const client = opcua.OPCUAClient.create({
-            endpointMustExist: false,
-            requestedSessionTimeout: 60000
+    it("server createSession should expose endpoints with correct productURI", async () => {
+        const client = OPCUAClient.create({});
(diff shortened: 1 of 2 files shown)


packages/node-opcua-end2end-test/test/end_to_end/test_e2e_1170.ts
```diff
@@ -0,0 +1,79 @@
...
+        let endpoints = server._get_endpoints();
+        endpointUrl = endpoints[0].endpointUrl;
+
+
+    });
+    after(function(done) {
+        server.shutdown(done);
+    });
+
+    it("should productURI have the same value", (done) => {
```
erossignon: Please favor new style with async/await instead of callback ... code will be easier to read.

packages/node-opcua-end2end-test/test/end_to_end/test_e2e_1170.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    const port = 1170;
+    let server = null;
+    let endpointUrl = "";
+
+
+    before(async () => {
+
+
+        server = new opcua.OPCUAServer({
```
erossignon, line 14: I would provide an explicit productUri in the OPCServer constructor, to make sure we verify against this one and not against the default value.

packages/node-opcua-end2end-test/test/end_to_end/test_e2e_1170.ts
```diff
@@ -0,0 +1,79 @@
...
+const opcua = require("node-opcua");
+const OPCUAClient = opcua.OPCUAClient;
+const UserTokenPolicy = opcua.UserTokenPolicy;
+const { perform_operation_on_client_session } = require("../../test_helpers/perform_operation_on_client_session");
+// require("node-opcua-service-session").UserNameIdentityToken;
+
+const describe = require("node-opcua-leak-detector").describeWithLeakDetector;
+describe("Testing bug #1170", function() {
+
+    const port = 1170;
```
erossignon, line 10: Please make sure that port is not used anywhere else in other unit test ... We run unit tests in parallel and all severs must have a different port.

packages/node-opcua-server/source/opcua_server.ts
```diff
@@ -262,9 +262,9 @@ async function _attempt_to_close_some_old_unactivated_session(server: OPCUAServe
 
 function getRequiredEndpointInfo(endpoint: EndpointDescription) {
     assert(endpoint instanceof EndpointDescription);
-    // It is recommended that Servers only include the server.applicationUri,  endpointUrl, securityMode,
-    // securityPolicyUri, userIdentityTokens, transportProfileUri and securityLevel with all
-    // other parameters set to null. Only the recommended parameters shall be verified by
+    // It is recommended that Servers only include the server.applicationUri, server.applicationType, server.applicationName,
```
erossignon: This sentences come from the OPCA spec ( probably v102). 
Now that the spec are available online we could add this link to the comment for cross reference : https://reference.opcfoundation.org/v104/Core/docs/Part4/5.6.2/

> List of Endpoints that the Server supports.
The Server shall return a set of EndpointDescriptions available for the serverUri specified in the request. The EndpointDescription type is defined in [7.10](https://reference.opcfoundation.org/v104/Core/docs/Part4/7.10). The Client shall verify this list with the list from a DiscoveryEndpoint if it used a DiscoveryEndpoint to fetch the EndpointDescriptions.
It is recommended that Servers only include the server.applicationUri, endpointUrl, securityMode, securityPolicyUri, userIdentityTokens, transportProfileUri and securityLevel with all other parameters set to null. Only the recommended parameters shall be verified by the client.


https://reference.opcfoundation.org/v105/Core/docs/Part4/5.6.2/
>List of Endpoints that the Server supports.  
> The Server shall return a set of EndpointDescriptions available for the serverUri specified in the request. All Endpoints are returned if the serverUri is null or empty. The EndpointDescription type is defined in [7.14](https://reference.opcfoundation.org/v105/Core/docs/Part4/7.14). The Client shall verify this list with the list from a DiscoveryEndpoint if it used a DiscoveryEndpoint to fetch the EndpointDescriptions.
It is recommended that Servers only include the server.applicationUri, endpointUrl, securityMode, securityPolicyUri, userIdentityTokens, transportProfileUri and securityLevel with all other parameters set to null or empty. Only the recommended parameters shall be verified by the Client.

erossignon (changes requested):

packages/node-opcua-end2end-test/test/end_to_end/test_e2e_1170.ts
JeanArhancet: This port is not used by other tests

packages/node-opcua-server/source/opcua_server.ts
JeanArhancet: Done @erossignon , What do you think?

packages/node-opcua-end2end-test/test/end_to_end/test_e2e_1170.ts
JeanArhancet: I added the parameter

packages/node-opcua-end2end-test/test/end_to_end/test_e2e_1170.ts
JeanArhancet: Thanks, I just changed

packages/node-opcua-end2end-test/test/end_to_end/test_e2e_1170.ts
```diff
@@ -0,0 +1,49 @@
...
+                productUri: "Mini NodeOPCUA-Server"
+            }
+        });
+
+        await server.start();
+
+        let endpoints = server._get_endpoints();
+        endpointUrl = endpoints[0].endpointUrl;
+    });
+    after(function (done) {
```
erossignon: I suggest this for consistency:
```
after(async ()=>{
   await server.shutdown();
});
```

packages/node-opcua-end2end-test/test/end_to_end/test_e2e_1170.ts
```diff
@@ -0,0 +1,49 @@
...
+        server.shutdown(done);
+    });
+
+    beforeEach(async () => {
```
erossignon: As we have only one test, there is no need for this beforeEach/afterEach block, we could also eliminate the variables at the describe scope

````javascript
it("server createSession should expose endpoints with correct productURI ", async()=> {
   const client = OPCUAClient.create({});
    const serverEndpoints = await client.withSessionAsync(async (session)=>{
            return session.serverEndpoints;
    });
    serverEndpoints[0].server.productUri).should.eql("Mini NodeOPCUA-Server");
}};

````

File: packages/node-opcua-end2end-test/test/end_to_end/test_e2e_1170.ts [hunk 7]
```diff
@@ -0,0 +1,49 @@
...
+        server = new opcua.OPCUAServer({
+            port,
+            serverInfo: {
+                productUri: "Mini NodeOPCUA-Server"
+            }
+        });
+
+        await server.start();
+
+        let endpoints = server._get_endpoints();
```
erossignon: let use the public API,
````
server.getEndpointUri();
```

erossignon (changes requested):

packages/node-opcua-end2end-test/test/end_to_end/test_e2e_1170.ts
JeanArhancet: You're right, I've changed

Issue #1215, opened 2022-11-11
UAMultiStateValueDiscrete value of UINT64 type results in exception


Graos:
NODE-OPCUA-SERVER stack version : 2.82.0
NODEJS version : v18.12.0

While loading the nodeset file (containing UAMultiStateValueDiscrete value of type UINT64 i.e i=9) in the node-opcua server stack results in the following exception

************************************************************************************************************************
**\node_modules\node-opcua-address-space\dist\src\data_access\ua_multistate_value_discrete_impl.js:157
            throw new Error("Invalid DataType in UAMultiStateValueDiscrete =>" + this.dataType.toString());
                  ^

Error: Invalid DataType in UAMultiStateValueDiscrete =>ns=0;i=9
    at UAMultiStateValueDiscreteImpl._post_initialize (\node_modules\node-opcua-address-space\dist\src\data_access\ua_multistate_value_discrete_impl.js:157:19)
    at Object.promoteToMultiStateValueDiscrete [as promoter] (\node_modules\node-opcua-address-space\dist\src\data_access\ua_multistate_value_discrete_impl.js:170:10)
    at \node_modules\node-opcua-address-space\dist\source\loader\namespace_post_step.js:42:34
    at Generator.next (<anonymous>)
    at \node_modules\node-opcua-address-space\dist\source\loader\namespace_post_step.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (\node_modules\node-opcua-address-space\dist\source\loader\namespace_post_step.js:4:12)
    at promoteObjectAndVariablesInNamespace (\node_modules\node-opcua-address-space\dist\source\loader\namespace_post_step.js:29:12)
    at \node_modules\node-opcua-address-space\dist\source\loader\namespace_post_step.js:57:13
    at Generator.next (<anonymous>)**

****************************************************************************************************************************
The issue is in the code snippet in ua_multistate_value_discrete_impl.js where in the MultiStateValueDiscrete  UINT64 value results in index value 0 which results in exception being thrown since the condition is  indexOf(this.dataType.value) <= 0

_post_initialize() {
        // MultiStateValueDiscrete Variables can have any numeric Data Type;
        // this includes signed and unsigned integers from 8 to 64 Bit length.
        // istanbul ignore next
        if (typeof this.dataType.value !== "number" ||
            [
                node_opcua_variant_1.DataType.UInt64,
                node_opcua_variant_1.DataType.Int64,
[...]

PR #1246, opened 2023-02-24
fix issue #1215

basyskom-kraemer:
Hi!

I spend a little time and try to fix the issue #1215.

The problem is, that the "indexOf" on line 188 in "ua_multistate_value_discrete_impl.ts" returns a zero by UInt64 datatype.
I delete the "=" and now it works fine.  :-)

Best regards from Germany
Andreas (basyskom GmbH)

### DIFF
--- a/packages/node-opcua-address-space/src/data_access/ua_multistate_value_discrete_impl.ts
+++ b/packages/node-opcua-address-space/src/data_access/ua_multistate_value_discrete_impl.ts
@@ -185,7 +185,7 @@ export class UAMultiStateValueDiscreteImpl<T, DT extends DataType>
                 DataType.Byte,
                 DataType.SByte,
                 26 /*Number*/
-            ].indexOf(this.dataType.value as number) <= 0
+            ].indexOf(this.dataType.value as number) < 0
         ) {
             throw new Error("Invalid DataType in UAMultiStateValueDiscrete =>" + this.dataType.toString());
         }


erossignon:
Thank you, 
Can you provide the associated  unit test  ?

basyskom-kraemer:
Hi!
Sorry for the question. But I don't understand, what you mean with unit test?
I only change the "=" in code, test it under Windows and Ubuntu with Node 14.x and 16.x on my local PC and everything works fine... Can you please tell me the steps I have to do? Thank you. :-)

AndreasHeine:
just a test that it does not return zero anymore ;)
a test what it should not return (expect to fail) and a test what it should return (expect to pass)

erossignon:
> Hi! Sorry for the question. But I don't understand, what you mean with unit test? I only change the "=" in code, test it under Windows and Ubuntu with Node 14.x and 16.x on my local PC and everything works fine... Can you please tell me the steps I have to do? Thank you. :-)

A bug fix  requires a non-regression test in order to be accepted.

- you need to provide a unit test with your code that demonstrates that your fix ( how small it could be) actually fixes the bug.
- the test you provide should faild if the fix is not present and should pass when the fix is present.
- you can look at the test forlder of the modules folder to see how unit test can be written.

erossignon:
superceeded by #1252 , closing

PR #1252, opened 2023-03-12
issue-1215-bugfix: UAMultiStateValueDiscrete value of UINT64 type results in exception

iliareshetov:
I have extracted data type validation into separate function and added full test coverage


iliareshetov:
Please close related issue : https://github.com/node-opcua/node-opcua/issues/1215.
@erossignon

PR #1260, opened 2023-03-28
Closes #802 improves the uninstallAlarmMonitoring function

### LINKED ISSUE #802: uninstallAlarmMonitoring: Have null check for all variables
Hi,

For safety use, it's better to have null check for all variables which are used in this function.

```
if (!_sessionPriv.$clientAlarmList) {
    return;
}
```
↓
```
if (!_sessionPriv.$clientAlarmList
     || !_sessionPriv.$monitoredItemForAlarmList
     || !_sessionPriv.$subscriptionforAlarmList ) {
    return;
}
```

https://github.com/node-opcua/node-opcua/blob/master/packages/node-opcua-client/source/alarms_and_conditions/client_alarm_tools.ts#L45

iliareshetov:
Closes #802
This pull request modifies the existing **uninstallAlarmMonitoring** function to improve the safety and reliability of the uninstallation process. The function now uses a new utility function, **areClientSessionPrivFieldsValid**, to check if all required fields in the **ClientSessionPriv** interface are valid before proceeding with the uninstallation process.

### DIFF
--- a/packages/node-opcua-client/source/alarms_and_conditions/client_alarm_tools.ts
+++ b/packages/node-opcua-client/source/alarms_and_conditions/client_alarm_tools.ts
@@ -24,7 +24,7 @@ function r(_key: string, o: { dataType?: unknown; value?: unknown }) {
     }
     return o;
 }
-interface ClientSessionPriv extends ClientSession {
+export interface ClientSessionPriv extends ClientSession {
     $clientAlarmList: ClientAlarmList | null;
     $monitoredItemForAlarmList: ClientMonitoredItem | null;
     $subscriptionforAlarmList: ClientSubscription | null;
@@ -32,7 +32,7 @@ interface ClientSessionPriv extends ClientSession {
 // ------------------------------------------------------------------------------------------------------------------------------
 export async function uninstallAlarmMonitoring(session: ClientSession): Promise<void> {
     const _sessionPriv = session as ClientSessionPriv;
-    if (!_sessionPriv.$clientAlarmList) {
+    if (!areClientSessionPrivFieldsValid(_sessionPriv)) {
         return;
     }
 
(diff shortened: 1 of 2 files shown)


iliareshetov:
Hello @erossignon Would it be possible for you to take a moment to review this PR? I understand your time is valuable, but it would be greatly appreciated if you could help us address this minor bug. Thank you for your consideration.

packages/node-opcua-client/source/alarms_and_conditions/client_alarm_tools.ts
```diff
@@ -24,15 +24,15 @@ function r(_key: string, o: { dataType?: unknown; value?: unknown }) {
     }
     return o;
 }
-interface ClientSessionPriv extends ClientSession {
+export interface ClientSessionPriv extends ClientSession {
```
erossignon, line 27: ClientSessionPriv  is private and should not be "exported"

packages/node-opcua-client/source/alarms_and_conditions/client_alarm_tools.ts
```diff
@@ -45,6 +45,18 @@ export async function uninstallAlarmMonitoring(session: ClientSession): Promise<
     return;
 }
 
+/**
+ * Checks if all fields in the `ClientSessionPriv` interface are valid.
+ * @param session - The `ClientSessionPriv` object to check.
+ * @returns `true` if all fields are valid, `false` otherwise.
+ */
+export function areClientSessionPrivFieldsValid(session: ClientSessionPriv): boolean {
```
erossignon, line 53: This should not be a exported function. 

The function is probably not needed either as it could be incorporated easily in line 35

packages/node-opcua-client/test/test_client_alarm.ts
```diff
@@ -77,3 +90,55 @@ describe("Testing client alarm", () => {
         alarmList.length.should.eql(2);
     });
 });
+
+describe("Ensure that all fields in the ClientSessionPriv interface are valid", () => {
```
erossignon, line 94: I am not sure the test is relevant here.
It doesn't highlight the faulty behavior we want to fix and only test a trivial internal function areClientSessionPrivFieldsValid that has been introduce in the PR. It should actually stressed the behavior described in #802

Can we provide a test that actually reproduce the bug that causes this PR to appear instead  ?

erossignon (changes requested):

PR #1269, opened 2023-04-13
Precise clock in async refresh

erossignon:


erossignon:
That is what PreciseClock is doing !

jacobq:
> That is what PreciseClock is doing !, implementation predates the use of BigInt in nodejs

Ah, gotcha. Sorry, I only looked at the changeset and grepped for hrtime -- now I see it over in https://github.com/node-opcua/node-opcua/blob/4f72684ec46fcfe7a097d0193081624c0e17fe28/packages/node-opcua-utils/source/hrtime.ts

PR #1270, opened 2023-04-13
Clean up nodeset definition/generation code

jacobq:
This change attempts to centralize information about official/companion nodesets in `node-opcua-nodesets` so that there is only one place that needs to be updated in order to add additional nodesets (vs. having to also manually update `node-opcua-convert-nodeset-to-javascript` and `packages/tsconfig.json` in addition to `node-opcua-nodesets`).

I noticed that not all `*.NodeSet2.xml` node sets contain type information, which surprised me, but I presume this is the correct behavior. (Has to do with whether or not the nodeset defines new types?)
To make this more obvious, output like the following now appears when generating type information:

```
There is no type information to generate for eumabois
There is no type information to generate for iolinkIODD
There is no type information to generate for irdi
Ignoring node-opcua-nodeset-eumabois since package folder does not exist (node-opcua/packages/node-opcua-nodeset-eumabois)
Ignoring node-opcua-nodeset-io-link-iodd since package folder does not exist (node-opcua/packages/node-opcua-nodeset-io-link-iodd)
Ignoring node-opcua-nodeset-irdi since package folder does not exist (node-opcua/packages/node-opcua-nodeset-irdi)
```


packages/node-opcua-nodesets/source/index.ts
```diff
@@ -25,88 +66,61 @@ export function constructNodesetFilename(filename: string) {
...
-// {{ --------------------------------------------------------------
-//   this block is use to force pkg to import nodesets in package
-path.join(__dirname, "nodesets/Opc.Ua.NodeSet2.xml");
-path.join(__dirname, "nodesets/Opc.Ua.Di.NodeSet2.xml");
-path.join(__dirname, "nodesets/Opc.Ua.Adi.NodeSet2.xml");
-path.join(__dirname, "nodesets/Opc.Ua.AutoID.NodeSet2.xml");
-path.join(__dirname, "nodesets/Opc.Ua.MachineVision.NodeSet2.xml");
-path.join(__dirname, "nodesets/Opc.Ua.Robotics.NodeSet2.xml");
-path.join(__dirname, "nodeset/Opc.Ua.Ijt.Tightening.NodeSet2.xml");
-// ------------------------------------------------------------- }}
```
jacobq, line 37: Does this really still do something? Looked like some kind of artifact from old packager or something to me. e.g. there is a typo on line 36 (missing `s` in `nodesets/`) so that line presumably does not serve any useful function.

erossignon: Yes this block is use to force pkg to import nodesets in package, when webpacking , 
Typo should be corrected

erossignon:
> There is no type information to generate for iolinkIODD
Well spotted, some are just place holder for dynamically generated type. (see IOLink companion spec for instance)

packages/node-opcua-nodesets/source/index.ts
jacobq: Interesting...so webpack detects the calls to `path.join`? Will it still work if we do it programmatically? e.g.,

```ts
// {{ --------------------------------------------------------------
//   this block is use to force pkg to import nodesets in package
allNodesetMeta.forEach((meta) => path.join(__dirname, `nodesets/${meta.xmlFile}`));
// ------------------------------------------------------------- }}
```

Do the braces/comments matter?

jacobq: Shouldn't it get handled automatically by this?
https://github.com/node-opcua/node-opcua/blob/master/webpackDLL.config.js#L64-L69

It's not obvious to me how webpack gets involved, e.g., I don't [see it listed as a dependency](https://github.com/node-opcua/node-opcua/search?q=webpack). I tried running:
```
$ pnpm -w add -D webpack-cli@4 webpack@4 webpack-node-externals babel-minify-webpack-plugin babili-webpack-plugin copy-webpack-plugin@4
$ pnpm install && pnpm run build && pnpm run pretest
$ webpack
```
from the root of the project (downgraded to Node v16.x to avoid openssl-related issue that seemed to come up). That produced this:

<details><summary>Webpack output (HEAD=be0507247906798a0f9e304a586ffda80e912308)</summary>

```
Hash: 13f554279b5b29344767
Version: webpack 4.46.0
Child
    Hash: 13f554279b5b29344767
    Time: 14136ms
    Built at: 04/13/2023 5:00:05 PM
                          Asset      Size  Chunks                   Chunk Names
        simple_server.bundle.js  2.65 MiB       0  [emitted]        simple_server
    simple_server.bundle.js.map  7.87 MiB       0  [emitted] [dev]  simple_server
    Entrypoint simple_server = simple_server.bundle.js simple_server.bundle.js.map
      [0] ./packages/node-opcua-assert/dist/index.js 978 bytes {0} [built]
      [1] ./packages/node-opcua-data-model/dist/index.js 1.52 KiB {0} [built]
      [7] external "chalk" 42 bytes {0} [built]
     [34] external "fs" 42 bytes {0} [built]
     [45] external "path" 42 bytes {0} [built]
     [54] external "os" 42 bytes {0} [built]
    [136] external "env-paths" 42 bytes {0} [built]
    [227] ./packages/node-opcua-samples/bin/simple_server.js 20.9 KiB {0} [built]
    [228] external "assert" 42 bytes {0} [built]
    [229] external "yargs/yargs" 42 bytes {0} [built]
    [230] external "node-opcua" 42 bytes {0} [built]
    [231] ./packages/node-opcua-vendor-diagnostic/dist/index.js 851 bytes {0} [built]
    [232] ./packages/node-opcua-vendor-diagnostic/dist/vendor_diagnostic_nodes.js 9.45 KiB {0} [built]
    [969] ./packages/node-opcua-address-space-for-conformance-testing/dist/index.js 865 bytes {0} [built]
    [970] ./packages/node-opcua-address-space-for-conformance-testing/dist/address_space_for_conformance_testing.js 53.4 KiB {0} [built]
        + 966 hidden modules
    
    WARNING in ./packages/node-opcua-vendor-diagnostic/dist/vendor_diagnostic_nodes.js 54:16-37
    Critical dependency: the request of a dependency is an expression
     @ ./packages/node-opcua-vendor-diagnostic/dist/index.js
     @ ./packages/node-opcua-samples/bin/simple_server.js
    
    WARNING in configuration
    The 'mode' option has not been set, webpack will fallback to 'production' for this value. Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
    You can also set it to 'none' to disable any default behavior. Learn more: https://webpack.js.org/configuration/mode/
```

</details>

Then I remove the special (comment-marked) block and ran again. Still appears to work and generate `dist/simple_server.bundle.js`, which is only a few bytes smaller. I ran [`js-beautify`](https://www.npmjs.com/package/js-beautify) on both bundles and the only diff was the removal of the lines calling `path.join`:

```diff
51622c51622
<         t.constructNodesetFilename = r, a.join(e, "nodesets/Opc.Ua.NodeSet2.xml"), a.join(e, "nodesets/Opc.Ua.Di.NodeSet2.xml"), a.join(e, "nodesets/Opc.Ua.Adi.NodeSet2.xml"), a.join(e, "nodesets/Opc.Ua.AutoID.NodeSet2.xml"), a.join(e, "nodesets/Opc.Ua.MachineVision.NodeSet2.xml"), a.join(e, "nodesets/Opc.Ua.Robotics.NodeSet2.xml"), a.join(e, "nodeset/Opc.Ua.Ijt.Tightening.NodeSet2.xml");
---
>         t.constructNodesetFilename = r;
```

What am I missing?

GoetzGoerisch: The Eumanbois NodeSet is just an example how to use Woodworking. 
This is not normative.
Many Companion Spec groups do so. e.g. Machine Tools.

jacobq: > The Eumanbois NodeSet is just an example how to use Woodworking. This is not normative. Many Companion Spec groups do so. e.g. Machine Tools.

Thanks @GoetzGoerisch, that makes sense (Etinenne also [explained below](https://github.com/node-opcua/node-opcua/pull/1270#issuecomment-1507582140)). Regarding this webpack-related block of code though, can you help me understand how it works? It looks like it does nothing (just generates some path strings that are never used). Supposedly it affects the packager (webpack@4?), but I do not know how to test this, so I am not sure if I should remove this block, leave it as-is (except correcting the 1 typo I saw), update it with a longer static list, or update it with a dynamic/programmatically-generated list. I'm new to this project so am probably misunderstanding something, but I think it would make more sense to have packaging-related concerns handled in packager configuration rather than by special/magic things added to the source code.

erossignon:
THank you @jacobq ! Great addition !
I have merged the code manually to fix a few merge issues and also iterated once more on the refactoring.

PR #1273, opened 2023-04-26
fix: replace instanceof Error with util.types.isNativeError

skrtheboss:
Use util.type.isNativeError in favor of instanceof Error to ensure that the check works in every environment. See https://nodejs.org/api/all.html#all_util_utiltypesisnativeerrorvalue.

I have encountered this issue while testing my code with jest. I discovered that this [check](https://github.com/node-opcua/node-opcua/blob/e143ff72418bb3db8c0a2cb8d4b7e54a90521a73/packages/node-opcua-transport/source/client_tcp_transport.ts#L258) always fails in jest environment. Related jest [issue](https://github.com/jestjs/jest/issues/2549).


erossignon:
merged manually, due to conflit resolution needed.

PR #1292, opened 2023-08-22
NEW: ServerOptions: CreateMissingNode-Method

thedigitaloffice:
Optional method in OPCUA-Server options passed through to server-subscription, called if a requested node is not configured.
Opens the ability to create tag-nodes dynamically according the client request.


erossignon:
@thedigitaloffice  thank you for your contribution, I like the idea of dynamic creation of nodes

Unfortunatly, this PR doesn't come with the mandatory unit tests. 

Also, we need to make sure that the node creation not only applies when a monitored item is created but also when the node is read, writing or browse so that the server behalf like if the node has always b een in the address space regardless of how it is accessed.

Is this something you could look into ?

erossignon:
Closing due to lack of activity.

PR #1314, opened 2023-10-12
feat: bind hostname to tcp server when passed to OPCUAServer

### LINKED ISSUE #1303: Node-opcua does not seem to honor hostname parameter passed in when creating TCP server.
**Current behavior / Describe the bug**

When creating an OPCUAServer with a certain hostname, node-opcua will instead create a TCP server with host 0.0.0.0. Not honoring the `hostname` parameter in the `OPCUAServer`.

**Step by step instruction to reproduce**

1. Copy files from gist
2. `npm install`
3. `node test.js`
4. Do a `netstat -an` in a cmd prompt
5. See that the application with the port 9999 has local address of 0.0.0.0

**Expected behavior**

I expect the OPCUAServer to honor the hostname being passed in, so if I specify 127.0.0.1 it should listen on that hostname.

**Screenshots**

image

**Gist**
\
gist: https://gist.github.com/tetanw/8697eb6bc7f0356d93b7b1f42715f306

**Suggested solution**

Pass the `hostname` to the `OPCUAServerEndPointclass` here https://github.com/node-opcua/node-opcua/blob/master/packages/node-opcua-server/source/opcua_server.ts#L3499. And then use it in the TCP listen method here https://github.com/node-opcua/node-opcua/blob/master/packages/node-opcua-server/source/server_end_point.ts#L513.

**Context**

- ( ) my request is related to node-opcua acting as an OPCUA CLIENT
- (x) my request is related to node-opcua acting as an OPCUA SERVER

- ( ) I have installed node-opcua from source ( using git clone)
- (x) I have installed node-opcua as a package ( using npm install )
- ( ) I am using an application that uses node-opcua

    - ( ) node-red
    - ( ) other : please specify

- Device: Lenovo Thinkstation P700
- OS version:

    - (x) Windows : version : Windows 10 Enterprise, 21H2
    - ( ) Linux : version : ______
    - ( ) MacOs : version : ______
    - ( ) Raspbian: version : ______
    - ( ) Other : specify : ______
   
- Description of the other OPCUA system I am trying to connect to: NA

- **node-opcua version:** 2.113.0

  
  

- **Node:**  v18.16.0

iliareshetov:
Closes #1303

Ensure that the OPCUAServer correctly binds to the provided hostname when creating the TCP server. 
Previously, it was defaulting to [unspecified IPv6 address](https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address) (::) when IPv6 is available, or the [unspecified IPv4 address](https://en.wikipedia.org/wiki/0.0.0.0) (0.0.0.0) otherwise.

## Testing
- Added new test cases to validate the hostname parameter behavior.

### DIFF
--- a/packages/node-opcua-server/source/opcua_server.ts
+++ b/packages/node-opcua-server/source/opcua_server.ts
@@ -720,7 +720,14 @@ export interface OPCUAServerEndpointOptions {
      * @default getFullyQualifiedDomainName()
      */
     hostname?: string;
-
+    /**
+     * Host IP address or hostname where the TCP server listens for connections.
+     * If omitted, defaults to listening on all network interfaces:
+     * - Unspecified IPv6 address (::) if IPv6 is available,
+     * - Unspecified IPv4 address (0.0.0.0) otherwise.
+     * Use this to bind the server to a specific interface or IP.
+     */
+    host?: string;
     /**
      * the TCP port to listen to.
      * @default 26543
@@ -1202,7 +1209,7 @@ export class OPCUAServer extends OPCUABaseServer {
                 endpointDefinitions.push({
                     port: options.port === undefined ? 26543 : options.port,
                     hostname: options.hostname || hostname,
-
+                    host: options.host,
                     allowAnonymous: options.allowAnonymous,
                     alternateHostname: options.alternateHostname,
                     disableDiscovery: options.disableDiscovery,
@@ -3537,12 +3544,12 @@ export class OPCUAServer extends OPCUABaseServer {
 
     private createEndpoint(
         port1: number,
-        serverOptions: { defaultSecureTokenLifetime?: number; timeout?: number }
+        serverOptions: { defaultSecureTokenLifetime?: number; timeout?: number; host?:string }
     ): OPCUAServerEndPoint {
         // add the tcp/ip endpoint with no security
         const endPoint = new OPCUAServerEndPoint({
             port: port1,
-
+            host: serverOptions.host,
             certificateManager: this.serverCertificateManager,
 
             certificateChain: this.getCertificateChain(),
--- a/packages/node-opcua-server/source/server_end_point.ts
+++ b/packages/node-opcua-server/source/server_end_point.ts
@@ -122,6 +122,10 @@ export interface OPCUAServerEndPointOptions {
      */
     port: number;
     /**
+     * the tcp host
+     */
+    host?: string;
+    /**
      * the DER certificate chain
      */
     certificateChain: Certificate;
@@ -208,6 +212,7 @@ export class OPCUAServerEndPoint extends EventEmitter implements ServerSecureCha
      * the tcp port
      */
     public port: number;
+    public host: string | undefined;
     public certificateManager: OPCUACertificateManager;
     public defaultSecureTokenLifetime: number;
     public maxConnections: number;
@@ -245,6 +250,7 @@ export class OPCUAServerEndPoint extends EventEmitter implements ServerSecureCha
         options.port = options.port || 0;
 
         this.port = parseInt(options.port.toString(), 10);
+        this.host = options.host;
         assert(typeof this.port === "number");
 
         this._certificateChain = options.certificateChain;
@@ -510,8 +516,13 @@ export class OPCUAServerEndPoint extends EventEmitter implements ServerSecureCha
             debugLog("server is listening");
         });
 
+        const listenOptions: net.ListenOptions = {
+            port: this.port,
+            host: this.host
+        };
+
         this._server!.listen(
-            this.port,
+            listenOptions,
             /*"::",*/ (err?: Error) => {
                 // 'listening' listener
                 debugLog(chalk.green.bold("LISTENING TO PORT "), this.port, "err  ", err);
--- a/packages/node-opcua-server/test/test_server_issue_1303.ts
+++ b/packages/node-opcua-server/test/test_server_issue_1303.ts
@@ -1,0 +1,128 @@
+import net from "net";
+import os from "os";
+import should from "should";
+import { OPCUAServer } from "..";
+
+async function findAvailablePort(): Promise<number> {
+    return new Promise((resolve) => {
+        const server = net.createServer();
+        server.listen(0);
+        server.on("listening", function () {
+            const port = (server.address() as net.AddressInfo).port;
+            console.log(`INFO: Found available port ${port}`);
+            server.close(() => resolve(port));
+        });
+    });
+}
+
+function findLanIp(): string | undefined {
+    const networkInterfaces = os.networkInterfaces();
+
+    for (const iface of Object.values(networkInterfaces)) {
+        if (!iface) {
+            continue;
+        }
+
+        for (const alias of iface) {
+            if (alias.family === "IPv4" && !alias.internal) {
+                return alias.address;
+            }
+        }
+    }
+
+    return undefined;
+}
+
+function checkServer(host: string | undefined, port: number): Promise<boolean> {
+    return new Promise((resolve) => {
+        const socket = net.createConnection(port, host, () => {
+            socket.end();
+            resolve(true);
+        });
+
+        socket.on("error", () => {
+            resolve(false);
+        });
+    });
+}
+
+async function testServerStartupAndShutdown(host: string | undefined, port: number, endpointsToTest: string[]): Promise<boolean[]> {
+    try {
+        const serverOptions = host === undefined ? { port } : { host, port };
+        console.log("INFO: starting OPCUAServer with serverOptions: ", serverOptions);
+        console.log("INFO: endpointsToTest: ", endpointsToTest);
+
+        const server = new OPCUAServer(serverOptions);
+        await server.start();
+
+        const results: boolean[] = [];
+
+        for (const endpoint of endpointsToTest) {
+            const isEndpointConnected = await checkServer(endpoint, port);
+            results.push(isEndpointConnected);
+        }
+
+        await server.shutdown();
+        server.dispose();
+
+        return results;
+    } catch (err) {
+        console.error(`ERROR: failed server startup/shutdown with msg: ${err.message}`);
+        throw err;
+    }
+}
+
+describe("OPCUAServer - issue#1303", () => {
+    it("should start a net.Server on loopback address", async () => {
+        const port = await findAvailablePort();
+        const host = "localhost";
+
+        const lanIP = findLanIp();
+        const endpointsToTest = ["localhost"];
+
+        console.log(lanIP ? `WARNING: LAN IP found: ${lanIP}` : "WARNING: No LAN IP available, skipping test...");
+
+        if (lanIP) {
+            endpointsToTest.push(lanIP);
+
+            const results = await testServerStartupAndShutdown(host, port, endpointsToTest);
+
+            should(results[0]).eql(true, "It should be possible to connect to loopback interface (localhost)");
+            should(results[1]).eql(false, "It should not be possible to connect to external LAN IP");
+        }
+    });
+
+    it("should start a net.Server on LAN IP", async () => {
+        const port = await findAvailablePort();
+        const host = findLanIp();
+
+        console.log(host ? `WARNING: LAN IP found: ${host}` : "WARNING: No LAN IP available, skipping test...");
+
+        if (host) {
+            const endpointsToTest = [host, "localhost"];
+            const results = await testServerStartupAndShutdown(host, port, endpointsToTest);
+
+            should(results[0]).eql(true, "It should be possible to connect to LAN IP");
+            should(results[1]).eql(false, "It should not be possible to connect to loopback interface (localhost)");
+        }
+    });
+
+    it("should start a net.Server on default host (ensures backward compatibility)", async () => {
+        const port = await findAvailablePort();
+        const lanIP = findLanIp();
+        const endpointsToTest = ["127.0.0.1", "localhost", "::1"];
+
+        console.log(lanIP ? `WARNING: LAN IP found: ${lanIP}` : "WARNING: No LAN IP available, skipping test...");
+
+        if (lanIP) {
+            endpointsToTest.push(lanIP);
+
+            const results = await testServerStartupAndShutdown(undefined, port, endpointsToTest);
+
+            should(results[0]).eql(true, "It should be possible to connect to loopback interface (127.0.0.1)");
+            should(results[1]).eql(true, "It should be possible to connect to loopback interface (localhost)");
+            should(results[2]).eql(true, "It should be possible to connect to loopback interface (::1)");
+            should(results[3]).eql(true, "It should be possible to connect to external LAN IP");
+        }
+    });
+});


iliareshetov:
> 
>   1643 passing (14m)
>   12 pending
>   1 failing
> 
>   1) testing Client - Umbrella-B 
>        SetTriggering
>          SetTriggering-3 it should return BadNothingToDo if both linksToAdd and linksToRemove are empty:
> 
>       AssertionError: must  have received a changed notification and one empty notif
>       + expected - actual
> 
>       -3
>       +2
>       
>       at Context.<anonymous> (node-opcua-end2end-test\test\end_to_end\u_test_e2e_set_triggering.ts:329:51)
>       at processTicksAndRejections (node:internal/process/task_queues:95:5)
>       at async Context.ff (node-opcua-leak-detector\src\resource_leak_detector.js:426:17)
> 
> 
> 
>  ELIFECYCLE  Test failed. See above for more details.
> Error: Process completed with exit code 1.

Hi, could someone suggest why this error keep occasionally happens only on windows CI ?

erossignon:
Thank you, Iliya,

I think that what you are trying to do is restrict the TCP port to listen through one of IP interfaces available on the computer on which the server is running.

At this time the server listen to IP 0.0.0.0 meaning all available IP addresses . 

However,  with the proposed change, the server will be assigned to only one interface the one that resolved to the hostname. 

In most cases hostname routes to 127.0.0.1, meaning that the server would not be reachable from the external facing IP of the computer. 

I would like a change that maintain by default the current behavior, to avoid support nightmare.

Instead, I would be in favor to add an addition optional `host` parameter ( different to hostname )  on the end point constructor, to specify on which 

If host is not provided , it defaults to 0.0.0.0 or IPV6 equivalent.
If host is specified , it could be the external facing IP of the computer

packages/node-opcua-server/source/opcua_server.ts
```diff
@@ -3498,12 +3498,13 @@ export class OPCUAServer extends OPCUABaseServer {
 
     private createEndpoint(
         port1: number,
+        hostname: string | undefined,
```
erossignon: use a `host` parameter instead, to distinguish from "hostname", host will be used to identify on which external or internal facing network interface available on the computer we want the socket to be listening. if not specified it would default to 0.0.0.0 meaning that the socket will listen on all available network interface, which is also the current behavior.

packages/node-opcua-server/source/opcua_server.ts
```diff
@@ -3528,6 +3529,7 @@ export class OPCUAServer extends OPCUABaseServer {
             throw new Error("internal error");
         }
         const hostname = getFullyQualifiedDomainName();
+        const serverOptionHostname = serverOption.hostname;
```
erossignon: serverOptionHostname must be distinguished from serverOption.hostname

packages/node-opcua-server/source/server_end_point.ts
```diff
@@ -245,6 +277,9 @@ export class OPCUAServerEndPoint extends EventEmitter implements ServerSecureCha
 
         this.port = parseInt(options.port.toString(), 10);
         assert(typeof this.port === "number");
+        if (options.hostname && options.hostname !== getFullyQualifiedDomainName()) {
+            this.initializeHostname(options.hostname);
```
erossignon: This await method will not resolve in time.

If we use a separate optional host parameter, this is not needed anymore

packages/node-opcua-server/source/server_end_point.ts
```diff
@@ -274,6 +309,10 @@ export class OPCUAServerEndPoint extends EventEmitter implements ServerSecureCha
         assert(this.serverInfo !== null && typeof this.serverInfo === "object");
     }
 
+    private async initializeHostname(desiredHostname: string) {
```
erossignon: Not needed

packages/node-opcua-server/source/server_end_point.ts
```diff
@@ -510,8 +548,16 @@ export class OPCUAServerEndPoint extends EventEmitter implements ServerSecureCha
             debugLog("server is listening");
         });
 
+        const listenOptions: net.ListenOptions = {
+            port: this.port
+        };
+
+        if (this.hostname !== undefined && typeof this.hostname === "string" && this.hostname.length > 0) {
```
erossignon: don't use hostname for the reason explained before

packages/node-opcua-server/test/test_server_issue_1303.ts
```diff
@@ -0,0 +1,102 @@
...
+                resolve(port);
+            });
+            server.close();
+        });
+    });
+}
+
+async function executeNetstat(hostnames: string[], port: number): Promise<void> {
+    return new Promise((resolve, reject) => {
+        const netstatProcess = spawn("sh", ["-c", "netstat -an"]);
```
erossignon: This test will not work on Windows. Make sure that we are platform agnostic

packages/node-opcua-server/test/test_server_issue_1303.ts
```diff
@@ -0,0 +1,102 @@
...
+            }
+        });
+    });
+}
+
+describe("OPCUAServer - issue#1303", () => {
+    it("should start a OPCUAServer on specific host and port", async () => {
+        try {
+            const port = await findAvailablePort();
+            const hostname = "127.0.0.1";
```
erossignon: OPCUA specification favors hostname to be 'real hostname'. 
Ideally oPCUA server always serve the localhost (127.0.0.1) plus one or more external facing network interface. 
The purpose of the change is to restrict an OPCUA server to be exposed to the selected external network interface when two or more exists on the computer.

packages/node-opcua-server/source/opcua_server.ts
i-mantegna: image

Hi Etienne,

Thank you for providing such a detailed description of the problem. 
I implemented the changes you suggested.

packages/node-opcua-server/source/server_end_point.ts
```diff
@@ -245,6 +305,9 @@ export class OPCUAServerEndPoint extends EventEmitter implements ServerSecureCha
 
         this.port = parseInt(options.port.toString(), 10);
         assert(typeof this.port === "number");
+        if (options.host !== undefined) {
+            this.initializeHost(options.host);
```
erossignon: As we are using a separated `host` parameter different from the `hostname` there is no need to call `initializeHost` any more, as we can assume that the variable will contain the IP address of hostname that resolve to the network interface that we want to use already.

initializeHost not been used, this will simplify this commit a lot, 

No need for validateHost and isLoopback address anymore.

packages/node-opcua-server/source/server_end_point.ts
```diff
@@ -274,6 +337,10 @@ export class OPCUAServerEndPoint extends EventEmitter implements ServerSecureCha
         assert(this.serverInfo !== null && typeof this.serverInfo === "object");
     }
 
+    private async initializeHost(host: string) {
```
erossignon: see above, we can remove this

packages/node-opcua-server/source/server_end_point.ts
```diff
@@ -510,8 +576,16 @@ export class OPCUAServerEndPoint extends EventEmitter implements ServerSecureCha
             debugLog("server is listening");
         });
 
+        const listenOptions: net.ListenOptions = {
+            port: this.port
+        };
+
+        if (this.host !== undefined && typeof this.host === "string" && this.host.length > 0) {
```
erossignon: host is string | undefined ,  undefined  resolves to false,  "" resolves to false too, therefore, we can simplify the code like this 
```
 if (this.host) { 
    listenOptions.host = this.host;
 }
```




which simplify further to

```
listenOptions.host = this.host;
``` 


which  can lead to 

```
        const listenOptions: net.ListenOptions = {
            port: this.port,
            host: this.host
        };
````

File: packages/node-opcua-server/test/test_server_issue_1303.ts [hunk 11]
```diff
@@ -0,0 +1,157 @@
...
+        }
+    });
+
+    it("should start a net.Server on default host (ensures backward compatibility)", async () => {
+        const port = await findAvailablePort();
+        await testServerStartupAndShutdown(undefined, port, ["0.0.0.0", "[::]", "::"]);
+    });
+});
+
+describe("isLoopbackAddress Function Tests", () => {
```
erossignon: We won't need this test anymore if isLoopbackAddress  not needed

erossignon (changes requested):
Illya, 
I have added a few comments io further simplify the code.

packages/node-opcua-server/source/server_end_point.ts
erossignon: This call is not needed anymore (see above)

iliareshetov:
I'm thinking about squashing the commits, but last time I did that, the build didn't start up. 
Any ideas on how to kick off the build manually? 
Or should I just leave everything as it is for now?

packages/node-opcua-server/test/test_server_issue_1303.ts
```diff
@@ -0,0 +1,87 @@
...
+    } catch (err) {
+        should.fail(err, undefined, `Error in server startup/shutdown: ${err.message}`);
+    }
+}
+
+describe("OPCUAServer - issue#1303", () => {
+    it("should start a net.Server on loopback address", async () => {
+        const port = await findAvailablePort();
+        const host = "127.0.0.1";
+        await testServerStartupAndShutdown(host, port);
```
erossignon: We may want to verify here that the server is reachable on the 127.0.0.1 loopback interface and not on the ip.

packages/node-opcua-server/test/test_server_issue_1303.ts
```diff
@@ -0,0 +1,87 @@
...
+    });
+
+    it("should start a net.Server on LAN IP", async () => {
+        const port = await findAvailablePort();
+        const host = findLanIp();
+        console.log(host ? `INFO: LAN IP found: ${host}` : "WARNING: No LAN IP available, skiping test...");
```
erossignon: We may want to verify here that the server is reachable on the LAN ip and not on the LAN loopback interface.

packages/node-opcua-server/test/test_server_issue_1303.ts
```diff
@@ -0,0 +1,87 @@
...
+        if (host) {
+            await testServerStartupAndShutdown(host, port);
+        }
+    });
+
+    it("should start a net.Server on default host (ensures backward compatibility)", async () => {
+        const port = await findAvailablePort();
+        await testServerStartupAndShutdown(undefined, port);
```
erossignon: we need to verify that the server is reachable on both LAN IP and Looback IP

erossignon:
> I'm thinking about squashing the commits, but last time I did that, the build didn't start up. Any ideas on how to kick off the build manually? Or should I just leave everything as it is for now?



This is how I do usually to update a PR and reorganize the commits by recombining them, or reordering them. 
I like doing this using the git command directly in a bash shell.


Look at your history
```
$ git log --format=short
```

```console
Author: iliareshetov <42715397+iliareshetov@users.noreply.github.com>
Date:   Wed Nov 15 19:33:20 2023 +0200

    #1303 refactor: update test cases

commit 012727a29f25dc0a1a874bba6caffd76ce5d4a07
Author: iliareshetov <42715397+iliareshetov@users.noreply.github.com>
Date:   Wed Nov 15 18:25:33 2023 +0200

    #1303 refactor: remove unnecessary host validation

commit da8d8f1efdda609b50ed92d4e4e4ff394036eb08
Author: iliareshetov <42715397+iliareshetov@users.noreply.github.com>
Date:   Wed Nov 15 15:54:17 2023 +0200

    #1303 refactor:  Add 'host' parameter to OPCUAServer for specific interface binding
    
    Introduced a new 'host' parameter in OPCUAServer to allow starting the net.Server on a specified network interface. Defaults to '0.0.0.0', enabling listening on all available interfaces for backward compatibility.

commit eab82aa78414bbe6bdfbd122a0a6073e3629db83
Author: iliareshetov <42715397+iliareshetov@users.noreply.github.com>
Date:   Thu Oct 12 16:22:49 2023 +0300

    #1303 : pass hostname to tcp endpoint

commit fa55b1c57d53d795c4dcb6c83698dbfe7a8ffe1e
Author: iliareshetov <42715397+iliareshetov@users.noreply.github.com>
Date:   Thu Oct 12 15:44:55 2023 +0300

    #1303 : add failing test case

commit 9e5a34a490fa82e7f3caed317e20df0cebc362aa
Author: Etienne Rossignon <etienne.rossignon@sterfive.com>
Date:   Fri Oct 6 20:56:19 2023 +0200

    improve tests coverage for validate_data_type_correctness
```

use git rebase interactive by specifying the SHA-1 of the commit just before yours
```
$git rebase -i 9e5a34a490fa82e7f3caed317e20df0cebc362aa
```


This will be shown in the editor

```
pick fa55b1c57 #1303 : add failing test case
pick eab82aa78 #1303 : pass hostname to tcp endpoint
pick da8d8f1ef #1303 refactor:  Add 'host' parameter to OPCUAServer for specific interface binding
pick 012727a29 #1303 refactor: remove unnecessary host validation
pick ed6867486 #1303 refactor: update test cases

# Rebase 9e5a34a49..ed6867486 onto 9e5a34a49 (5 commands)
```

now replace `pick`with `squash except for the first line

```
pick fa55b1c57 #1303 : add failing test case
squash eab82aa78 #1303 : pass hostname to tcp endpoint
squash da8d8f1ef #1303 refactor:  Add 'host' parameter to OPCUAServer for specific interface binding
squash 012727a29 #1303 refactor: remove unnecessary host validation
squash ed6867486 #1303 refactor: update test cases

# Rebase 9e5a34a49..ed6867486 onto 9e5a34a49 (5 commands)
```

and exit the editor,

git will take over and squash the commit for you

then a window will appear to let you update the comment of the combined commit

Mkae sure that your commit comment do not start with a # sign, you may have to 
replace `#1303` with  `issue #1303` because # is used as a comment line in this file and git will be confused.


```
$ git rebase -i 9e5a34a490fa82e7f3caed317e20df0cebc362aa
[detached HEAD 5e3640194]  isssue #1303 refactor:  Add 'host' parameter to OPCUAServer for specific interface binding
 Author: iliareshetov <42715397+iliareshetov@users.noreply.github.com>
 Date: Thu Oct 12 15:44:55 2023 +0300
 3 files changed, 111 insertions(+), 7 deletions(-)
 create mode 100644 packages/node-opcua-server/test/test_server_issue_1303.ts
Successfully rebased and updated refs/heads/add-hostname-to-tcp-listen-options.
```

then 

```
$ git log --format=short
```

to see the result

You'll need to git force-push  to rewrite the history on your PR branch on github

```
$ git push -f 
```

packages/node-opcua-server/test/test_server_issue_1303.ts
```diff
@@ -0,0 +1,87 @@
...
+        const server = new OPCUAServer(serverOptions);
+        await server.start();
+
+        const isServerListening = await checkServer(host, port);
+        isServerListening.should.be.true;
+
+        await server.shutdown();
+        server.dispose();
+    } catch (err) {
+        should.fail(err, undefined, `Error in server startup/shutdown: ${err.message}`);
```
erossignon: `should.fail(err) will silently ignore when the code in the try block will raise an exception. 

For instance if  isServerListening.should.be.true; fails, then we will never catch this.

packages/node-opcua-server/test/test_server_issue_1303.ts
erossignon: May be we could adapt the async function testServerStartupAndShutdown function by passing an additional array with the endpoint to test. 
We could also return array with the connection status true/false for each endpoint above. 
Then the actual test would be done inside the it("...",()=>{}); method to facilitate the understanding of what the test is actually doing. 

```
async function testServerStartupAndShutdown(host: string | undefined, port: number, endpointsToTest: string[]) {
  ...
  // start server with host and port
  // for each endpointsToTest   
  //      check it can be connected 
  //       store the result in the array
  return the array with result 
}

```

```
it(should start a net.Server on loopback address", async () => {
    const port = await findAvailablePort();
    const host = "localhost";

    const lanIP = findLanIp();
    const endpontsToTest = [
       "127.0.0.1",
       lanIP,
    ];

    const result = await testServerStartupAndShutdown(host, port, endpointsToTest);

    should(result[0]).eql(true, "It should be possible to connect to loopback interface");
    should(result[1]).eql(false, "It should not be possible to connect to external lan IP");
 
}
```

You get the idea

iliareshetov:
Hi, is there anything else that needs to be done?

erossignon:
Hi Ilia,
Yes, there are still pending review issues that need to be resolved, in particular in the unit test.

erossignon (changes requested):

iliareshetov:
Thank you for your feedback; I'll make the necessary improvements.
I couldn't view your comments even though you left them three weeks ago.

I've only been able to view the comments after you explicitly requested changes.

erossignon:
Yes, my bad, I forgot to submit the review,

Issue #1326, opened 2023-12-08
Node-opcua inherits variables from objectType with same name as variable browseName.


tetanw:
**Current behavior / Describe the bug**

When a variable's BrowseName on one objectType matches the name of another objectType that inherits from BaseObjectType then it will take over the reference from that objectType when instantiating.

See the configuration in the Siemens modelling editor:
![image](https://github.com/node-opcua/node-opcua/assets/6298455/fc573181-de5f-4401-82dd-02c4dca1fc0d)
![image](https://github.com/node-opcua/node-opcua/assets/6298455/787e76d0-9d3b-438f-99e9-0681e2d8234c)

See how the instantiated node looks like:
![image](https://github.com/node-opcua/node-opcua/assets/6298455/571fcbce-1412-4b3a-a166-51ce98007004)

**Step by step instruction to reproduce**

1. Copy files from gist into a folder
2. Copy the standard nodeset `Opc.Ua.NodeSet2.xml` into the folder as well
3. `npm install`
4. `node server.js`
5. Open any OpcUa browser and see the issue

**Expected behavior**

In the example scenario in the image I would expect only the Name variable to be there.

**Gist**

gist: https://gist.github.com/tetanw/248223774efafe9a5593dbd55f063145

**Context**

- ( ) my request is related to node-opcua acting as an OPCUA CLIENT
- (x) my request is related to node-opcua acting as an OPCUA SERVER

- ( ) I have installed node-opcua from source ( using git clone)
- (x) I have installed node-opcua as a package ( using npm install )
- ( ) I am using an application that uses node-opcua

    - ( ) node-red
    - ( ) other : please specify

- Device: Lenovo Thinkstation P700
- OS version:

    - (x) Windows : version : Windows 10 Enterprise, 21H2
    - ( ) Linux : version : ______
    - ( ) MacOs : version : ______
    - ( ) Raspbian: version : ______
    - ( ) Other : specify : ______
   
- Description of the other OPCUA system I am trying to connect to: NA

- **node-opcua version:** 2.118.0

  
  

- **Node:**  v18.16.0

tetanw:
Might actually be related to #1320.

tetanw:
Just to clarify this only goes wrong when doing dynamic object instantiation, i.e. `objectType.instantiate(...)`. When I add a xml file with the objectType already instantiated by the Siemens OPCUA modelling editor, this just works correctly.

tetanw:
[...]

PR #1328, opened 2023-12-11
issue #1326 bug-fix: no longer adding nodes from an unrelated object …

tetanw:
Bug fix implementing suggestion from #1326.


erossignon:
already merged

PR #1343, opened 2024-02-26
fix displayName or description do not exist in object when crawling

narttmk:
For some reason, displayName or description do not exist in object when crawling so it throw an exception about can not get text from null or undefined.


erossignon:
now merged into master

PR #1376, opened 2024-06-26
fixing issue#1375 wrong dataType for minSupportedSampleRate

erossignon:


packages/node-opcua-server/source/server_capabilities.ts
```diff
@@ -64,7 +64,7 @@ export interface IServerCapabilities {
     maxArrayLength: number;
     maxByteStringLength: number;
     maxQueryContinuationPoints: number;
-    minSupportedSampleRate: number;
+    minSupportedSampleRate: Double;
```
AndreasHeine, line 67: isn't it Duration?

erossignon: Duration = Double = number in javascript

PR #1407, opened 2025-01-10
Dispose MonitoredItem after event

patrickl92:
Prior to emitting the `removeMonitoredItem` event, the MonitoredItem instance is disposed. This resets all properties of that MonitoredItem. When the `removeMonitoredItem` event is raised, a subscriber cannot retrieve any information from this instance (i.e. the item which was monitored).
Therefore the MonitoredItem instance is disposed after emitting the event.


erossignon:
Great catch. May I ask you to create a dedicated unit test that will prevent potention future regression  and capture the expected behavior ?

patrickl92:
Added a unit test as requested

PR #1475, opened 2026-02-02
Feat/perf improvement

erossignon:


packages/node-opcua-end2end-test/test/end_to_end/test_e2e_ResultDataType_on_from_client.ts
```diff
@@ -0,0 +1,161 @@
...
+        dataType: DataType.ExtensionObject,
+        value: addressSpace.constructExtensionObject(threeDVectorType, {
+            x: 1.0,
+            y: 2.0,
+            z: 3.0
+        })
+    });
+
+    const extObj = addressSpace.constructExtensionObject(resultDataType, {
+        ResultMetadata: {
```
AndreasHeine, line 79: @erossignon the main problem occures if you use a Subtype of ResultMetaDataType here (which has AllowSubtypes=True) -> https://reference.opcfoundation.org/Machinery/Result/v101/docs/10.5

The client is unable to resolve the ResultMetadata (the Subtype) in an ExtensionObject which results in an Opaque Structure instead...

erossignon: This looks like a typo in the doc : see https://mantis.opcfoundation.org/view.php?id=10852

PR #1476, opened 2026-02-03
improve(feat): PushCertificateManagerServerImpl

AndreasHeine:
Hey @erossignon 

not sure if it is intendet to not install the issuer certs from the update method in server configuration object!?


packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -85,36 +157,39 @@ export interface PushCertificateManagerServerOptions {
...
 
 export async function moveFile(source: string, dest: string): Promise<void> {
-    debugLog("moving file file \n source ", source, "\n =>\n dest ", dest);
+    debugLog("moving file \n source ", source, "\n =>\n dest ", dest);
     try {
         await copyFile(source, dest);
         await deleteFile(source);
     } catch (err) {
-        errorLog(err);
+        errorLog("Error moving file from", source, "to", dest, ":", err);
```
erossignon: just display err.messsage this way, no the entire stack: (we have it in errorLog already)

errorLog("Error moving file from", source, "to", dest, ":", (err as Error).message);

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -85,36 +157,39 @@ export interface PushCertificateManagerServerOptions {
...
-        const exists = await fs.existsSync(file);
-        if (exists) {
-            debugLog("deleting file ", file);
+    const exists = fs.existsSync(file);
+    if (exists) {
+        debugLog("deleting file ", file);
+        try {
             await fs.promises.unlink(file);
+        } catch (err) {
+            errorLog("Error deleting file", file, ":", err);
```
erossignon: errorLog("Error deleting file", file, ":", (err as Error).message);

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -74,6 +67,85 @@ function findCertificateGroupName(certificateGroupNodeId: NodeId | string): stri
...
+        }
+        
+        // ECC algorithm OIDs: 1.2.840.10045.x
+        if (algorithm.startsWith("1.2.840.10045")) {
+            return "ECC";
+        }
+        
+        return null;
+    } catch (err) {
+        errorLog("Error extracting certificate key type:", err);
```
erossignon: errorLog("Error extracting certificate key type:", (err as Error).message);

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -85,36 +157,39 @@ export interface PushCertificateManagerServerOptions {
...
-        debugLog("copying file \n source ", source, "\n =>\n dest ", dest);
-        const sourceExist = fs.existsSync(source);
-        if (sourceExist) {
+    debugLog("copying file \n source ", source, "\n =>\n dest ", dest);
+    const sourceExist = fs.existsSync(source);
+    if (sourceExist) {
+        try {
             await fs.promises.copyFile(source, dest);
+        } catch (err) {
+            errorLog("Error copying file from", source, "to", dest, ":", err);
```
erossignon: errorLog("Error copying file from", source, "to", dest, ":", (err as Error).message);

packages/node-opcua-server-configuration/test/server/test_push_certificate_manager_server_impl.ts
```diff
@@ -274,4 +275,806 @@ describe("Testing Server Side PushCertificateManager", () => {
...
+
+        // Then the result should be Good
+        result.statusCode.should.eql(StatusCodes.Good);
+        result.certificateSigningRequest!.should.be.instanceOf(Buffer);
+
+        // And a temporary certificate manager should be created
+        (pushManager as any)._tmpCertificateManager.should.not.be.undefined();
+
+        // Ensure temporary certificate manager is cleaned up for subsequent tests
+        await pushManager.applyChanges("DefaultApplicationGroup");
```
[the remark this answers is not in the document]
AndreasHeine: hm interesting how did it passes test !?

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -206,6 +313,50 @@ export class PushCertificateManagerServerImpl extends EventEmitter implements Pu
...
+                        } catch (err) {
+                            warningLog(`Failed to remove stale pending file ${file}:`, err);
+                        }
+                    }
+                }
+            } catch (err) {
+                warningLog(`Error during pending files cleanup in ${issuerFolder}:`, err);
+            }
+        }
     }
```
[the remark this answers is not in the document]
AndreasHeine: fair point!

packages/node-opcua-server-configuration/test/server/test_push_certificate_manager_server_impl.ts
```diff
@@ -274,4 +275,806 @@ describe("Testing Server Side PushCertificateManager", () => {
...
+        } finally {
+            // Restore backed up certificates
+            if (fs.existsSync(issuerFolder)) {
+                const currentFiles = fs.readdirSync(issuerFolder);
+                for (const file of currentFiles) {
+                    if (file.startsWith("issuer_")) {
+                        fs.unlinkSync(path.join(issuerFolder, file));
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
AndreasHeine: fair point!

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -17,19 +17,12 @@ import {
     certificateMatchesPrivateKey,
     coercePEMorDerToPrivateKey,
     coercePrivateKeyPem,
+    DirectoryName
 } from "node-opcua-crypto/web";
 
 import { readPrivateKey, readCertificate } from "node-opcua-crypto";
```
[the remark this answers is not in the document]
AndreasHeine: not sure how to treat that... if someone imports the Interface from this file and i remove the export it would break existing stuff...

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -650,19 +886,26 @@ export class PushCertificateManagerServerImpl extends EventEmitter implements Pu
...
+            if (false) {
+                // node 10.2 and above
+                for await (const task of t) {
+                    await task();
+                }
+            } else {
+                while (t.length) {
+                    const task = t.shift()!;
+                    await task();
+                }
```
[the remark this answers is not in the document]
AndreasHeine: technical debt ^^

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -17,19 +17,12 @@ import {
...
-// fixme: use the one from node-opcua-crypto
-export interface DirectoryName {
-    stateOrProvinceName?: string;
-    localityName?: string;
-    organizationName?: string;
-    organizationUnitName?: string;
-    commonName?: string;
-    countryName?: string;
-}
+export { DirectoryName } // kept for backward compatibility
```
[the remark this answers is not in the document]
erossignon: No need to re-export, now DirectoryName  is in node-opcua-crypto/web other places shall use the import " node-opcua-crypto/web" to avoid confusion

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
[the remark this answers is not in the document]
erossignon: So fix the existing stuff

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -232,22 +454,122 @@ export class PushCertificateManagerServerImpl extends EventEmitter implements Pu
...
             const currentCertificateFilename = path.join(certificateManager.rootDir, "own/certs/certificate.pem");
-            if (!fs.existsSync(currentCertificateFilename)) {
-                errorLog("Cannot find existing certificate to extract subjectName", currentCertificateFilename);
+            try {
+                const certificate = readCertificate(currentCertificateFilename);
+                const e = exploreCertificate(certificate);
+                subjectName = subjectToString(e.tbsCertificate.subject);
+                warningLog("reusing existing certificate subjectName = ", subjectName);
+            } catch (err) {
+                errorLog("Cannot find existing certificate to extract subjectName", currentCertificateFilename, ":", err);
```
erossignon: (err as Error).message

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -258,20 +580,20 @@ export class PushCertificateManagerServerImpl extends EventEmitter implements Pu
...
 
             const location = path.join(certificateManager.rootDir, "tmp");
-            if (fs.existsSync(location)) {
-                await rimraf.rimraf(path.join(location));
-            }
-            if (!fs.existsSync(location)) {
-                await fs.promises.mkdir(location);
-            }
+            // Clean up existing tmp directory if present
+            // rimraf doesn't throw if directory doesn't exist
```
erossignon: No need for // rimraf doesn't throw if directory doesn't exist

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -258,20 +580,20 @@ export class PushCertificateManagerServerImpl extends EventEmitter implements Pu
...
+            await rimraf.rimraf(path.join(location));
+            // Create fresh tmp directory
+            // Using recursive: true is safe - doesn't throw if directory already exists
```
erossignon: No need for // Using recurisveursive: true is safe - doesn't throw if directory already exists

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -85,43 +235,52 @@ export interface PushCertificateManagerServerOptions {
...
-        debugLog("copying file \n source ", source, "\n =>\n dest ", dest);
-        const sourceExist = fs.existsSync(source);
-        if (sourceExist) {
-            await fs.promises.copyFile(source, dest);
-        }
+        await fs.promises.copyFile(source, dest);
     } catch (err) {
-        errorLog(err);
+        errorLog("Error copying file from", source, "to", dest, ":", err);
+        throw err;
```
erossignon: We don't want to rethrow here,

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -85,43 +235,52 @@ export interface PushCertificateManagerServerOptions {
...
 
 export async function copyFile(source: string, dest: string): Promise<void> {
+    debugLog("copying file \n source ", source, "\n =>\n dest ", dest);
     try {
-        debugLog("copying file \n source ", source, "\n =>\n dest ", dest);
-        const sourceExist = fs.existsSync(source);
-        if (sourceExist) {
-            await fs.promises.copyFile(source, dest);
-        }
+        await fs.promises.copyFile(source, dest);
```
erossignon: I am not sure copyFile(source, dest)  doesn't throuh if source doesn't exists .
I would keep as it was.

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -85,43 +235,52 @@ export interface PushCertificateManagerServerOptions {
...
     }
 }
 
 export async function moveFileWithBackup(source: string, dest: string): Promise<void> {
-    // let make a copy of the destination file
+    // let make a copy of the destination file if it exists
     debugLog("moveFileWithBackup file \n source ", source, "\n =>\n dest ", dest);
-    await copyFile(dest, dest + "_old");
+    try {
+        await fs.promises.access(dest);
```
erossignon: We should handle gracefully cases were dest doesn't not exist yet, this is normal operatio and this should not raise any exceptions.
try to avoid try { } catch and rethrow

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
AndreasHeine: @erossignon it was mainly because its a mix of existSync and and async copyFile operation now its all async!

erossignon: While doing this you changed the logic of the method. you need to restore it

AndreasHeine: OH! your right... let me fix that... 
but i would stick to all async without changing the logic!?

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
AndreasHeine: > try to avoid try { } catch and rethrow

in general (scope of the PR) or in that case?

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -56,22 +47,179 @@ const defaultApplicationGroup = resolveNodeId("ServerConfiguration_CertificateGr
...
+ * FIXME: Consider extracting certificateTypeId validation into a separate function
+ * that doesn't require the certificate, to avoid code duplication between
+ * createSigningRequest and updateCertificate.
+ * 
+ * @param certificate The certificate to validate
+ * @param certificateTypeId The NodeId of the expected certificate type
+ * @returns true if valid or if validation is not applicable
+ */
+function validateCertificateType(certificate: Buffer, certificateTypeId: NodeId | string): boolean {
+    // FIXME: Certificate type validation should check against CertificateTypes property
```
AndreasHeine: any thoughts?

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -232,22 +451,122 @@ export class PushCertificateManagerServerImpl extends EventEmitter implements Pu
...
         }
 
+        // OPC UA Part 12 - CreateSigningRequest Method:
+        // The certificateTypeId specifies the type of Certificate being requested.
+        // The set of permitted types is specified by the CertificateTypes Property 
+        // belonging to the Certificate Group. Valid types are defined in OPC UA Part 12, Section 6.3.2.
+        // Note: Unlike certificateGroupId, there is no "If null" clause in the spec for certificateTypeId.
+        // However, null NodeId and empty string are accepted (treated as "no type validation").
+        
+        // FIXME: Certificate type validation should check against CertificateTypes property
```
AndreasHeine: any thoughts?

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -443,8 +845,42 @@ export class PushCertificateManagerServerImpl extends EventEmitter implements Pu
 
         // The Server shall do all normal integrity checks on the Certificate and all of the issuer
         // Certificates. If errors occur the BadSecurityChecksFailed error is returned.
-        // todo : all normal integrity check on the certificate
-        const certInfo = exploreCertificate(certificate);
+        
+        // FIXME: Additional certificate validation checks required per OPC UA Part 4 & 6:
```
AndreasHeine: any thoughts?

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -56,22 +47,179 @@ const defaultApplicationGroup = resolveNodeId("ServerConfiguration_CertificateGr
...
     }
-    if (sameNodeId(certificateGroupNodeId, defaultUserTokenGroup)) {
+    if (sameNodeId(nodeId, defaultUserTokenGroup)) {
         return "DefaultUserTokenGroup";
     }
-    return "";
+    
+    // If it's a valid NodeId but not recognized, return empty string
+    // If it was originally a string (and not a standard group), return the string as group name
+    return typeof certificateGroupNodeId === "string" ? certificateGroupNodeId : "";
```
erossignon: This is wierd, Why this ?

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -56,22 +47,179 @@ const defaultApplicationGroup = resolveNodeId("ServerConfiguration_CertificateGr
...
+}
+
+/**
+ * Extract the key type from a certificate (RSA or ECC)
+ * @param certificate The certificate to analyze
+ * @returns "RSA" or "ECC" or null if unknown
+ */
+function getCertificateKeyType(certificate: Buffer): "RSA" | "ECC" | null {
```
erossignon: THis method belongs to node-opcua-crypto

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
erossignon: We have something that does this alerady in the securechannel area, same checks applies, Rather than duplicating  checks we need to carefully refactor and reuse. 
I would to this in a second PR.

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -508,18 +967,33 @@ export class PushCertificateManagerServerImpl extends EventEmitter implements Pu
             // a new certificate is provided for us,
             // we keep our private key
             // we do this in two stages
-            await preInstallCertificate(this);
+            try {
+                await preInstallIssuerCertificates(this, baseCounterValue);
+                await preInstallCertificate(this, baseCounterValue + (issuerCertCount * 2), baseCounterValue + (issuerCertCount * 2) + 1);
```
erossignon: What is this magic formula ?

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -540,29 +1015,42 @@ export class PushCertificateManagerServerImpl extends EventEmitter implements Pu
...
                 warningLog("certificate doesn't match privateKey");
                 return { statusCode: StatusCodes.BadSecurityChecksFailed, applyChangesRequired: false };
             }
 
-            await preInstallPrivateKey(this);
-
-            await preInstallCertificate(this);
-
-          
+            try {
```
erossignon: Why would preInstallCertficate fails  and throw, normal operation should be that it doesnt't throw.

erossignon: We need to separate preinstallation from verification

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -610,6 +1098,18 @@ export class PushCertificateManagerServerImpl extends EventEmitter implements Pu
         // If the Server Certificate has changed, Secure Channels using the old Certificate will
         // eventually be interrupted.
 
+        // Check if there are any pending tasks
+        if (this._pendingTasks.length === 0 && this.$$actionQueue.length === 0) {
+            // If ApplyChanges is called and there is no active transaction then return Bad_NothingToDo
+            return StatusCodes.BadNothingToDo;
```
erossignon: Good !

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -610,6 +1098,18 @@ export class PushCertificateManagerServerImpl extends EventEmitter implements Pu
...
+        }
+
+        // FIXME: Event handler error isolation
```
erossignon: Remove this comment, or implement the suggested fix

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -650,19 +1153,36 @@ export class PushCertificateManagerServerImpl extends EventEmitter implements Pu
...
-            // node 10.2 and above
-            for await (const task of t) {
-                await task();
-            }
-        } else {
-            while (t.length) {
-                const task = t.shift()!;
-                await task();
+        // FIXME: Transaction rollback capability
+        // If a task fails partway through, we currently clean up pending files but don't
```
erossignon: Interesting comment from the AI. 
Let's keep this in mind and implemenent in a nexgt PR

packages/node-opcua-server-configuration/source/push_certificate_manager.ts
```diff
@@ -64,13 +64,19 @@ export interface PushCertificateManager {
      * BadSecurityChecksFailed      Some failure occurred verifying the integrity of the Certificate.
      *
      */
+    updateCertificate(
+      certificateGroupId: NodeId | string,
+      certificateTypeId: NodeId | string,
+      certificate: ByteString,
+      issuerCertificates: ByteString[]
+    ): Promise<UpdateCertificateResult>;
     updateCertificate(
```
erossignon, line 66: So if you propose 2 flavors of updateCertificate , one should be with no private key and the other should be with private key stuff mandatory, not optionals

AndreasHeine: this might be a leftover... i remeber having ts error but after i made the pkey stuff optional... its obsolete... i will remove it!

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
AndreasHeine: i tried to avoid race condition if you have multiple call's at the same time the filnames would collide or am i am wrong here?

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
AndreasHeine: AI helps if you ask the right questions ;)
it surprised me in the level of assistence -> less boring typing, more thinking and creating but you need to be carefull context is the most important thing!

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
AndreasHeine: @erossignon do you have a better solution for that i mean its a rare case?

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
erossignon: We have Certficate validation already somewhere ( used for client<->server connection cerficicate checks).

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
erossignon: Yes in this PR , because the try { }  has just been introduced here ...

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
erossignon: I am afraid there is no async equivalent for fs.existsSync , you are allowed to use the sync version for checking file existence;

AndreasHeine: fs.promises.access:

Checks whether the process has specific permissions:
 F_OK  – existence (default)

 R_OK  – readable

 W_OK  – writable

 X_OK  – executable (makes sense on Unix; limited semantics on Windows)


It's a tradeoff between time-of-check and time-of-use, anyhow i thougth it would be suitable to so and handle possible Errors. Nevermind i Stick to your requirements and advice 😉

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
AndreasHeine: Ok, let me rework the Error Handling to use try/Catch only where i have to do cleanup on failure

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
AndreasHeine: there is an old existing test which would fail!

AndreasHeine: @erossignon 

Ok i revisited this one...

the function takes nodeid or string as input!

UA Spec says:
if GroupId is not provided (NullNodeId / EmptyNodeId) use DefaultApplicationGroup

empty string != not provided

resolveNodeId would return -> "ns=0;i=0" which would lead to return DefaultApplicationGroup !?

**BUT is an empty string the same as a not existing Group or is it the same as not provided ???**

so the possible return values are:
```
1. "" // Shit in Shit out ^^
2. "DefaultApplicationGroup" // default or valid group nodeid
3. "DefaultHttpsGroup" // valid group nodeid
4. "DefaultUserTokenGroup" // valid group nodeid
5. "MyCustomGroup" // Custom not found with resolve
6. "ns=5;i=1234" // unknown at least not a standard one so -> Shit in Shit out ^^
```

```ts
    private getCertificateManager(certificateGroupId: NodeId | string): CertificateManager | null {
        const groupName = findCertificateGroupName(certificateGroupId);
        return this._map[groupName] || null;
    }
```

getCertificateManager returns
```
1. "" -> null
2. "DefaultApplicationGroup" -> CertificateManager-Instance
3. "DefaultHttpsGroup" -> CertificateManager-Instance
4. "DefaultUserTokenGroup" -> CertificateManager-Instance
5. "MyCustomGroup" -> null
6. "ns=5;i=1234" -> null if not implemented !?
```

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
AndreasHeine: @erossignon in another PR?

Together with:
```
FIXME: Transaction rollback capability
```

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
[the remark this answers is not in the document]
erossignon: DirectoryName  is only used in push_cerficiate_manager_server module and was exported to be used by an other script of the same module, this is not a public API stuff that push_cerficiate_manager module should expose. 
Just remove the re-export statement.

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
erossignon: Yes, that's will be a new feature to add in a new PR.

packages/node-opcua-server-configuration/source/push_certificate_manager.ts
erossignon: No, having 2 overloads for the public API is good, one without private key and one with the two arguments required for when we want to use it with a private key. 
This enfore the user to use the right one and not forget to specify format when key is provided. 
Please fix by adding proper overloading .

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
erossignon: Yes, you need to properly handle simulatneaous calls of 

One way is to set a flag ( a mutex in effect) as soon as one caller make the first all,  and unset it when the entire operation is completed. 
If a second call happen once the first one in in progress,it will see the flag being set and will reject the operation with a StatusCode like : StatusCodes.BadInvalidOperation

erossignon: [ ] implement the mutex
[ ] remove the formula
[ ] remove the uncessary try/catch

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
erossignon: Not really, because this is a change in the code that has just been inserted, please do it in this PR.

erossignon:
> > We have far too many commits in this PR for just a single feature. Please consider squashing all your iterative commits into a single commit: This will help the possibility to backport the feature if need be.
> 
> @erossignon doesn't "squash and merge" did it anyways ?

THis will garbage the commit message with 60 unwanted message concatenation. 

It is cleaner to do it on your computer, craft a suitable commit message , then force push.

AndreasHeine:
@erossignon sorry for the delay, just to recap before i get back to work on it:

TODO / TOFIX:
- separate preinstallation from verification + Transaction rollback capability
- remove "magic formula" ^^ -> mutex
- FIXME: Additional certificate validation checks -> check secure channel!
- getCertificateKeyType -> check if exist in node-opcua-crypo if use the one from crypto else move there!?

NEW Issues:

- TrustList Write does not work due to invalid Open Mode:
https://github.com/node-opcua/node-opcua/blob/ab3bd98e69a7d365a40094d35122c4e879d04460/packages/node-opcua-server-configuration/source/server/promote_trust_list.ts#L150
https://github.com/node-opcua/node-opcua/blob/ab3bd98e69a7d365a40094d35122c4e879d04460/packages/node-opcua-server-configuration/source/clientTools/push_certificate_management_client.ts#L203

packages/node-opcua-server-configuration/source/server/promote_trust_list.ts
```diff
@@ -67,50 +67,120 @@ async function _closeAndUpdate(
...
 
-            // Update the certificate manager with new data
-            if (trustListData.trustedCertificates) {
-                for (const cert of trustListData.trustedCertificates) {
-                    await cm.trustCertificate(cert);
-                }
+            // Automatically update specifiedLists mask based on which fields are actually present
+            // This ensures that fields with data are processed even if the mask wasn't set correctly
+            if (trustListData.issuerCrls && trustListData.issuerCrls.length > 0) {
+                trustListData.specifiedLists |= 0x08; // IssuerCrls mask
```
erossignon: Use Mask enum not 0xYY numbers

packages/node-opcua-server-configuration/source/server/promote_trust_list.ts
```diff
@@ -67,50 +67,120 @@ async function _closeAndUpdate(
...
+            if (trustListData.trustedCertificates && trustListData.trustedCertificates.length > 0) {
+                trustListData.specifiedLists |= 0x01; // TrustedCertificates mask
+            }
+            if (trustListData.issuerCertificates && trustListData.issuerCertificates.length > 0) {
+                trustListData.specifiedLists |= 0x04; // IssuerCertificates mask
+            }
+
+            // Helper function to clear all CRL files from a directory
+            const clearCrlFolder = async (folder: string) => {
+                const fsModule = require("fs") as typeof import("fs");
```
erossignon: Use import on top of the file, not embeded requires

packages/node-opcua-server-configuration/source/server/promote_trust_list.ts
```diff
@@ -67,50 +67,120 @@ async function _closeAndUpdate(
...
+                const pathModule = require("path") as typeof import("path");
+                try {
+                    if (!fsModule.existsSync(folder)) {
```
erossignon: fs.existsSync

packages/node-opcua-server-configuration/source/server/promote_trust_list.ts
```diff
@@ -67,50 +67,120 @@ async function _closeAndUpdate(
...
-                const issuerCrlFolder = cm.issuersCrlFolder;
+            if (trustListData.trustedCertificates && trustListData.trustedCertificates.length > 0) {
+                trustListData.specifiedLists |= 0x01; // TrustedCertificates mask
+            }
+            if (trustListData.issuerCertificates && trustListData.issuerCertificates.length > 0) {
+                trustListData.specifiedLists |= 0x04; // IssuerCertificates mask
+            }
+
+            // Helper function to clear all CRL files from a directory
+            const clearCrlFolder = async (folder: string) => {
```
erossignon: Move this inner helper  function outside in the global scope

packages/node-opcua-server-configuration/source/server/promote_trust_list.ts
```diff
@@ -67,50 +67,120 @@ async function _closeAndUpdate(
...
+                            await fsModule.promises.unlink(filePath);
+                        }
+                    }
+                } catch (err) {
+                    warningLog("Error clearing CRL folder:", folder, err);
+                }
+            };
+
+            // Helper function to write CRL data directly to disk
+            const writeCrlToDisk = async (folder: string, crlBuffer: Buffer, index: number) => {
```
erossignon: Move this inner helper  function outside in the global scope

packages/node-opcua-server-configuration/source/server/promote_trust_list.ts
```diff
@@ -67,50 +67,120 @@ async function _closeAndUpdate(
...
+                const fsModule = require("fs") as typeof import("fs");
```
erossignon: no embeeded require, please

packages/node-opcua-server-configuration/source/server/promote_trust_list.ts
```diff
@@ -67,50 +67,120 @@ async function _closeAndUpdate(
...
+                }
+                
+                // Use .crl extension for the files
+                const filename = pathModule.join(folder, `crl_${index}.crl`);
+                await fsModule.promises.writeFile(filename, crlBuffer);
+            };
+
+            // Process CRLs based on specifiedLists to implement replace/update behavior
+            // Clear and replace issuer CRLs if they are part of the specified lists
+            if ((trustListData.specifiedLists & 0x08) === 0x08) { // IssuerCrls mask
```
erossignon: Use a flag enum instead of number for mask( same elsewhere)

packages/node-opcua-server-configuration/source/server/promote_trust_list.ts
```diff
@@ -67,50 +67,120 @@ async function _closeAndUpdate(
...
+                // Get or construct issuersCrlFolder path
+                const pathModule = require("path") as typeof import("path");
```
erossignon: same

packages/node-opcua-server-configuration/source/server/trust_list_server.ts
```diff
@@ -53,13 +54,49 @@ export async function buildTrustList(
         trustList.trustedCertificates = await readAll(certificateManager.trustedFolder);
     }
     if ((trustListFlag & TrustListMasks.TrustedCrls) === TrustListMasks.TrustedCrls) {
-        trustList.trustedCrls = await readAll(certificateManager.crlFolder);
+        // Get or construct crlFolder path
+        let crlFolder = (certificateManager as any).crlFolder;
```
erossignon: => remove any

packages/node-opcua-server-configuration/source/server/trust_list_server.ts
```diff
@@ -53,13 +54,49 @@ export async function buildTrustList(
...
+        if (crlFolder && fs.existsSync(crlFolder)) {
+            trustList.trustedCrls = await readAll(crlFolder);
+        } else {
+            trustList.trustedCrls = [];
+        }
     }
     if ((trustListFlag & TrustListMasks.IssuerCertificates) === TrustListMasks.IssuerCertificates) {
-        trustList.issuerCertificates = await readAll(certificateManager.issuersCertFolder);
+        // Get or construct issuersCertFolder path
+        let issuersCertFolder = (certificateManager as any).issuersCertFolder;
```
erossignon: no any, please

packages/node-opcua-server-configuration/source/server/trust_list_server.ts
```diff
@@ -53,13 +54,49 @@ export async function buildTrustList(
...
+        if (!issuersCertFolder) {
+            const rootDir = certificateManager.rootDir || (certificateManager as any).location;
```
erossignon: no any please

packages/node-opcua-server-configuration/source/server/trust_list_server.ts
```diff
@@ -53,13 +54,49 @@ export async function buildTrustList(
...
+        if (issuersCertFolder && fs.existsSync(issuersCertFolder)) {
+            trustList.issuerCertificates = await readAll(issuersCertFolder);
+        } else {
+            trustList.issuerCertificates = [];
+        }
     }
     if ((trustListFlag & TrustListMasks.IssuerCrls) === TrustListMasks.IssuerCrls) {
-        trustList.issuerCrls = await readAll(certificateManager.issuersCrlFolder);
+        // Get or construct issuersCrlFolder path
+        let issuersCrlFolder = (certificateManager as any).issuersCrlFolder;
```
erossignon: no any please

packages/node-opcua-server-configuration/test/test_server_configuration.ts
```diff
@@ -339,9 +339,12 @@ describe("ServerConfiguration", () => {
             newTrustList.trustedCertificates = a.trustedCertificates;
             newTrustList.issuerCertificates = a.issuerCertificates;
             newTrustList.trustedCrls = a.trustedCrls;
+            
+            // Use real CRL data from the certificate authority
+            const { crl } = await _getFakeAuthorityCertificate(await initializeHelpers("TEST_CRL", 0));
             newTrustList.issuerCrls = [
-                    Buffer.from("dummyCRL", "utf-8"),
-                    Buffer.from("dummyCRL2", "utf-8")
+                    crl,
```
erossignon: Add more unit tests for the CRL feature .to improve code coverage

AndreasHeine:
removeCertificate method in promote_trust_list.ts is not implemented.... hang on ^^

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -399,10 +499,59 @@ export class PushCertificateManagerServerImpl extends EventEmitter implements Pu
...
+                    await writeFile(pendingIssuerFilePEM, toPem(issuerCert, "CERTIFICATE"), "utf-8");
+                    
+                    // Stage the move operations to be applied on applyChanges()
+                    self.addPendingTask(() => moveFileWithBackup(pendingIssuerFileDER, finalIssuerFileDER));
+                    self.addPendingTask(() => moveFileWithBackup(pendingIssuerFilePEM, finalIssuerFilePEM));
+                    
+                    debugLog(`Staged issuer certificate ${i + 1}/${issuerCertificates.length}: ${thumbprint}`);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
AndreasHeine: but then, in case of a rollback, i dont know which one to remove or not.... now i only overwrite on apply changes!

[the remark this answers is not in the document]
AndreasHeine: would be cleanedup on restart -> class PushCertificateManagerServerImpl.initialize()!?

packages/node-opcua-server-configuration/source/server/push_certificate_manager_server_impl.ts
```diff
@@ -74,6 +67,117 @@ function findCertificateGroupName(certificateGroupNodeId: NodeId | string): stri
...
+            return "ECC";
+        }
+        
+        warningLog("Unknown certificate signature algorithm:", algorithmStr);
+        return null;
+    } catch (err) {
+        errorLog("Error extracting certificate key type:", err);
+        return null;
+    }
+}
```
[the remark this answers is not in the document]
AndreasHeine: i should consider that!

erossignon:
Thanks  @AndreasHeine 
Now fully integrated + more in master

PR #1480, opened 2026-02-09
fix(mdns): emit onRegisterServer events for mDNS discovered servers

AlexanderThees:
Fixes #1479

### 1. mDNS server discovery events
- `MDNSResponder` now extends `EventEmitter` and emits typed `serverUp`/`serverDown` events
- `OPCUADiscoveryServer` forwards these as unified `onRegisterServer`/`onUnregisterServer` events
- Discovery clients can now handle mDNS-discovered servers the same way as LDS-registered servers

### 2. Hostname extraction fix
- Added `parseEndpointUrl()` to extract hostname from endpoint URL
- Replaces `"TODO-find how to extract hostname"` placeholder
- Falls back to `getFullyQualifiedDomainName()` if parsing fails

## Files Changed

- `packages/node-opcua-server/source/register_server_manager_mdns_only.ts`
- `packages/node-opcua-server-discovery/source/mdns_responder.ts`
- `packages/node-opcua-server-discovery/source/opcua_discovery_server.ts`

### DIFF
--- a/packages/node-opcua-server/source/register_server_manager_mdns_only.ts
+++ b/packages/node-opcua-server/source/register_server_manager_mdns_only.ts
@@ -6,8 +6,23 @@
 import { EventEmitter } from "events";
 import { assert } from "node-opcua-assert";
 import { BonjourHolder } from "node-opcua-service-discovery";
+import { getFullyQualifiedDomainName } from "node-opcua-hostname";
 import { OPCUABaseServer } from "./base_server";
 import { IRegisterServerManager, RegisterServerManagerStatus } from "./i_register_server_manager";
+
+/**
+ * Extract hostname and path from an OPC UA endpoint URL
+ * @param endpointUrl - URL like "opc.tcp://hostname:port/path"
+ * @returns { host, path } or null if parsing fails
+ */
+function parseEndpointUrl(endpointUrl: string): { host: string; path: string } | null {
+    const match = endpointUrl.match(/^opc\.tcp:\/\/([^:/]+)(?::\d+)?(\/.*)?$/);
+    if (!match) return null;
+    return {
+        host: match[1],
+        path: match[2] || "/"
+    };
+}
 
 /**
  * a RegisterServerManager that declare the server the OPCUA Bonjour service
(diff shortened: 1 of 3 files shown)


packages/node-opcua-server/source/register_server_manager_mdns_only.ts
```diff
@@ -6,9 +6,24 @@
...
 import { OPCUABaseServer } from "./base_server";
 import { IRegisterServerManager, RegisterServerManagerStatus } from "./i_register_server_manager";
 
+/**
+ * Extract hostname and path from an OPC UA endpoint URL
+ * @param endpointUrl - URL like "opc.tcp://hostname:port/path"
+ * @returns { host, path } or null if parsing fails
+ */
+function parseEndpointUrl(endpointUrl: string): { host: string; path: string } | null {
+    const match = endpointUrl.match(/^opc\.tcp:\/\/([^:/]+)(?::\d+)?(\/.*)?$/);
```
erossignon, line 19: Please use the URL utility provided by NodeJS, rather than crafting a regexp

```
a = new URL("opc.tcp://hostname:1234/some/path")
URL {
  href: 'opc.tcp://hostname:1234/some/path',
  origin: 'null',
  protocol: 'opc.tcp:',
  username: '',
  password: '',
  host: 'hostname:1234',
  hostname: 'hostname',
  port: '1234',
  pathname: '/some/path',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

erossignon (changes requested):
Thank you Alexander,

Please also add additional unit tests to cover the new behavior and demonstrate correct update of DNS.? 
There should be plenty of unit test already to get inspiration

packages/node-opcua-server/source/register_server_manager_mdns_only.ts
AlexanderThees: Hi @erossignon,

thank you for the suggestion 👍
You're absolutely right — using NodeJS URL is cleaner and more robust than relying on a custom RegExp.

I'll refactor the implementation to use the built-in URL utility and will also add additional unit tests to cover the MDSN / LDS callbacks asap.

I’ll push an update shortly.

Thanks again for the review!

erossignon: @AlexanderThees , kind reminder. just to make sure your contribution will fit in the next incremental release.

Issue #1496, opened 2026-04-23
RFC OPC UA browser client via WebSocket Transport


kwojcicki:
# RFC: OPC UA browser client via WebSocket transport (`opc.ws://` / `opc.wss://`)

I've prototyped (and written some playwright tests) a browser-capable OPC UA client on top of `node-opcua` and I'd like your feedback/blessing on the approach before opening PRs.

The prototype works end-to-end against an unmodified `node-opcua-server` (fronted by a small TCP↔WebSocket bridge): The prototype validates use cases like: Read / Write / CreateSubscription / MonitoredItem notifications, over both `opc.ws://` with `SecurityPolicy.None` and `opc.wss://` with `Basic256Sha256` + `SignAndEncrypt` + `UserNameIdentityToken`.

The following is the list of PRs I would publish with general changes made

#### 1. `node-opcua-debug` — guard `__filename` on non-Node runtimes 
`make_loggers.ts` dereferences `__filename` inside `setDebugFlag`, which is undefined in ESM/browser contexts. One `typeof __filename === "undefined" ? "<browser>" : __filename` fallback.

#### 2. `node-opcua-transport` — lazy `os.hostname()` in `client_tcp_transport.ts`
The module currently calls `os.hostname()` at top level to build a default `serverUri`. In browser bundles where `os` is polyfilled-or-absent this throws at module load. Lazy-init inside a function, fallback to `"localhost"` if `os.hostname` isn't callable.

#### 3. `node-opcua-secure-channel` — minor runtime polyfills
- `client_secure_channel_layer.ts`: replace `setImmediate(...)` with `typeof setImmediate === "function" ? setImmediate : microtask-fallback` (one call site).
- `verify_pcks1.ts`: early-return on `typeof process === "undefined" || !process.version`. The function runs at module load and fails regex-parsing an empty `process.version` under the standard browser `process` polyfill.

#### 4. `node-opcua-transport` — `IClientTransport` interface + `IClientTransportFactory`

A new public interface describing the contract `ClientTCP_transport` already fulfils. 

`defaultClientTransportFactory` returns `new ClientTCP_transport(...)`.

I put it in `node-opcua-transport` because `ClientTCP_transport` lives there and the interface codifies its public surface. Easy to move if you prefer.

#### 5. `node-opcua-secure-channel` — optional `transportFactory` on `ClientSecureChannelLayerOptions` 

[...]

PR #1503, opened 2026-05-04
feat(secure-channel): add transportFactory option to ClientSecureChanelLayerOptions

kwojcicki:
Follow-up to the IClientTransport / IClientTransportFactory / defaultClientTransportFactory exports added in https://github.com/node-opcua/node-opcua/pull/1501. Wires the factory through ClientSecureChannelLayer so consumers can substitute the transport implementation without forking node-opcua-secure-channel.

Changes in client_secure_channel_layer.ts to swap interface and factory
New test `test/20-test_transport_factory_seam.ts`

PR 5 from https://github.com/node-opcua/node-opcua/issues/1496

### DIFF
--- a/packages/node-opcua-secure-channel/test/20-test_transport_factory_seam.ts
+++ b/packages/node-opcua-secure-channel/test/20-test_transport_factory_seam.ts
@@ -1,0 +1,138 @@
+import should from "should";
+import sinon from "sinon";
+
+import {
+    ClientTCP_transport,
+    defaultClientTransportFactory,
+    type IClientTransport,
+    type IClientTransportFactory,
+    type TransportSettingsOptions
+} from "node-opcua-transport";
+
+import { ClientSecureChannelLayer } from "../dist/source";
+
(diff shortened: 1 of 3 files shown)


packages/node-opcua-secure-channel/test/20-test_transport_factory_seam.ts
```diff
@@ -0,0 +1,138 @@
...
+                return stub as unknown as IClientTransport;
+            }
+        };
+        const createSpy = sinon.spy(factory, "create");
+
+        const channel = new ClientSecureChannelLayer({ transportFactory: factory });
+
+        // Kick off connect. Stub never completes the handshake, but that's fine:
+        // we only want to observe which factory path was taken.
+        (channel as any).create("fake://localhost:1/Fake", () => {
```
erossignon: avoid `as any` as we are moving soon to using biome with standard settings where `as any` is strongly discouraged.

PR #1504, opened 2026-05-04
feat(common): in-memory certificate provider for OPCUASecureObject

kwojcicki:
Adds an optional in-memory path for OPCUASecureObject to get its identity certificate chain and private key, bypassing the disk-backed readCertificateChain / readPrivateKey calls. This unblocks browser clients (which have no fs) and simplifies test fixtures that would otherwise have to stage PKI folders on disk.

PR 6 from https://github.com/node-opcua/node-opcua/issues/1496

### DIFF
--- a/packages/node-opcua-common/source/opcua_secure_object.ts
+++ b/packages/node-opcua-common/source/opcua_secure_object.ts
@@ -172,6 +172,24 @@ export function getPartialCertificateChain(certificateChain?: Certificate | Cert
 export interface IOPCUASecureObjectOptions {
     certificateFile?: string;
     privateKeyFile?: string;
+    /**
+     * Optional pre-built certificate + private-key provider. When supplied,
+     * `OPCUASecureObject.getCertificate()` / `.getCertificateChain()` /
+     * `.getPrivateKey()` delegate to this object verbatim, and the disk-backed
+     * `SecretHolder` path (`fs.existsSync` + `readCertificateChain` +
+     * `readPrivateKey`) is not used.
+     *
+     * Intended for browser builds (bundled via esbuild) and test fixtures that
+     * want to stage a cert+key pair without staging PKI folders on disk. A
+     * caller that wants in-memory semantics simply passes an implementation of
+     * {@link ICertificateKeyPairProvider} whose methods return values from
+     * memory.
+     *
+     * When present, `certificateFile` / `privateKeyFile` become optional and
+     * may be omitted (internal placeholders are used for toJSON output).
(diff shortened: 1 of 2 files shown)


packages/node-opcua-common/source/opcua_secure_object.ts
```diff
@@ -20,19 +20,44 @@ export interface IHasCertificateFile {
...
+ *
+ * Intended for browser builds (bundled via esbuild) and for test fixtures
+ * that want to stage a cert+key without staging PKI folders on disk.
+ *
+ * Node production callers generally should NOT use this; the disk-backed
+ * path supports self-signed-cert auto-generation, shared trust stores, and
+ * `fs.watch`-driven live reload, all of which an in-memory provider would
+ * have to re-implement.
+ */
+export interface IInMemoryCertificateProvider {
```
erossignon: IInMemoryCertificateProvider  is almost identical to ICertificateKeyPairProvider µ

I would rather inject a ICertificateKeyPairProvider that have a different behavior rather than adding an additional IInMemoryCertificateProvider optional argument taht makes the API a little bit wierd 

I think that it s worth doing now

kwojcicki: Just to make sure we are aligned you envision something like:

```typescript
export interface IOPCUASecureObjectOptions {
    certificateFile?: string;
    privateKeyFile?: string;
    certificateKeyPairProvider?: ICertificateKeyPairProvider; <--- new
}
```

Or are you proposing a 

```typescript
export class InMemoryOPCUASecureObject<T extends Record<string | symbol, any> = any>
    extends EventEmitter<T>
    implements ICertificateKeyPairProvider, IHasCertificateFile
{
```

That then clients can utilize instead of `OPCUASecureObject`? The reason I didn't go with the later is `OPCUASecureObject` is extended several times and swapping to composition would add a lot of boilerplate to expose the existing OPCUASecureObject methods.

erossignon: The idea is to push further the refactoring.
let me propose a improvement in the architecture that makes OPCUAClient, OPUCAServer not dependent on node:fs, node-opcua-pki, 
By cleaning up those unwanted dependencies first, we will make the in memory certificate provider concept easier to integrate.

erossignon: Your commit has been incorporated into #1505. This PR also contains the restructuring to move away reference to node:fs and node-opcua-pki

PR #1511, opened 2026-05-14
feat(client-browser): createBrowserClient() helper

kwojcicki:
Adds a one-call helper that constructs an `OPCUAClient` pre-wired with browser-appropriate defaults so consumers don't have to assemble the WebSocket transport factory + in-memory cert/key provider + in-memory trust store manually.


erossignon:
please rebase against origin/master to ensure ci is clean now

PR #1534, opened 2026-06-30
fix(server): drop oldest retransmission entry on overflow

MrAlaskan:
### Summary
This PR fixes retransmission queue overflow handling for sent `NotificationMessage`.

OPC UA Part 4 says that when a retransmission queue overflows, the oldest sent `NotificationMessage` gets deleted. The current implementation trims the queue from the wrong side and can remove newer messages instead of the oldest retained one.

### Bug
`discardOldSentNotifications()` currently uses `splice(length - maxNotificationMessagesInQueue)` once the queue reaches its limit. That does not remove the oldest sent message. Instead, it starts deleting entries from the tail of the array, which is where newer sent notifications are stored.

### Changes
- Only trim the retransmission queue when it actually exceeds the configured maximum
- Remove the excess entries from the head of the queue so that the oldest sent messages are discarded first
- Add a regression test covering the overflow case and verifying that newer sequence numbers remain available

### DIFF
--- a/packages/node-opcua-server/test/test_subscription.ts
+++ b/packages/node-opcua-server/test/test_subscription.ts
@@ -983,6 +983,44 @@ describe("Subscriptions", function (this: any) {
             // TODO
         });
 
+        it("should drop the oldest sent NotificationMessage when the retransmission queue overflows", () => {
+            const subscription = makeSubscription({
+                id: 1234,
+                publishingInterval: 1000,
+                lifeTimeCount: 1000,
+                maxKeepAliveCount: 20,
+                publishEngine: fake_publish_engine,
+                globalCounter: { totalMonitoredItemCount: 0 },
+                serverCapabilities: { maxMonitoredItems: 10000, maxMonitoredItemsPerSubscription: 1000 }
+            });
+
+            (subscription as any)._sent_notification_messages = Array.from({ length: 100 }, (_, index) => ({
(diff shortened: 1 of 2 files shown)


packages/node-opcua-server/test/test_subscription.ts
```diff
@@ -983,6 +983,44 @@ describe("Subscriptions", function (this: any) {
...
+                id: 1234,
+                publishingInterval: 1000,
+                lifeTimeCount: 1000,
+                maxKeepAliveCount: 20,
+                publishEngine: fake_publish_engine,
+                globalCounter: { totalMonitoredItemCount: 0 },
+                serverCapabilities: { maxMonitoredItems: 10000, maxMonitoredItemsPerSubscription: 1000 }
+            });
+
+            (subscription as any)._sent_notification_messages = Array.from({ length: 100 }, (_, index) => ({
```
erossignon: avoid 'as any', create a dedicated interface ISubscriptionPrive type that has a _sent_notification_messages method, 
Create  a subscriptionInternal objects = subscription as unknown as ISubscriptionPrive  and assert that _sent_notification_messages exists as a function (so we catch early if the inner implementation changes later).

MrAlaskan: Done, I replaced the `as any` cast with a dedicated internal test interface and added explicit assertions for the internal queue shape.

PR #1535, opened 2026-06-30
fix(file-transfer): preserve contents for write open modes

MrAlaskan:
### Summary
This change fixes the server-side mapping of OPC UA `FileType` open modes to Node.js file flags.

The `Write` and `ReadWrite` modes are meant to allow writes without implicitly erasing existing file contents. Only the `EraseExisting` bit is supposed to request truncation of the file during `Open`.

### Bug
`toNodeJSMode()` currently maps both `OpenFileMode.Write` and `OpenFileMode.ReadWrite` to `"w+"`.

In Node.js, `"w+"` truncates the file as soon as it is opened. This collapses the distinction between ordinary write access and the explicit `EraseExisting` modes, because `Open(Write)` and `Open(ReadWrite)` end up erasing the file even when the `EraseExisting` bit is not set.

As a result, a client can silently destroy existing file contents by performing a normal writable `Open`, which is inconsistent with the `FileType` open mode definition.

### Changes
Map `OpenFileMode.Write` and `OpenFileMode.ReadWrite` to a non-truncating readable/writable flag

### DIFF
--- a/packages/node-opcua-file-transfer/source/server/file_type_helpers.ts
+++ b/packages/node-opcua-file-transfer/source/server/file_type_helpers.ts
@@ -306,7 +306,7 @@ function toNodeJSMode(opcuaMode: OpenFileMode): string {
             break;
         case OpenFileMode.ReadWrite:
         case OpenFileMode.Write:
-            flags = "w+";
+            flags = "r+";
             break;
         case OpenFileMode.ReadWriteAppend:
         case OpenFileMode.WriteAppend:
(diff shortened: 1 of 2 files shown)


erossignon:
@MrAlaskan, some unit tests got broken and need to be fixed.

MrAlaskan:
Thanks for the reminder. I updated the unit tests, and now all checks pass except build_on_windows (24.15.0) .
I haven't been able to pinpoint the cause of that remaining failure yet.

PR #1536, opened 2026-06-30
fix(server): honor serverDiagnosticsEnabled option

MrAlaskan:
This PR fixes a typo in `ServerEngine` option handling. The constructor checked `serverDiagnosticsEnable` instead of `serverDiagnosticsEnabled`, which meant the explicit `serverDiagnosticsEnabled` option could be ignored and the engine would fall back to the default value. This change corrects the property name so the configured diagnostics flag is applied as intended.


packages/node-opcua-server/source/server_engine.ts
```diff
@@ -507,7 +507,7 @@ export class ServerEngine extends EventEmitter implements IAddressSpaceAccessor
             this._applicationUri = options.applicationUri || "<unset _applicationUri>";
         }
 
-        options.serverDiagnosticsEnabled = Object.hasOwn(options, "serverDiagnosticsEnable")
+        options.serverDiagnosticsEnabled = Object.hasOwn(options, "serverDiagnosticsEnabled")
```
erossignon: I propose an further enhancement instead
why not :
    options.serverDiagnosticsEnabled == undefined ? true : options.serverDiagnosticsEnabled

Object.hasOwn is not type aware → risk of misspelling the property, and only your IA finds out ;-)

erossignon (changes requested):

packages/node-opcua-server/source/server_engine.ts
MrAlaskan: Haha, true, AI can be surprisingly handy for code auditing. I closed this PR and opened a fresh one instead (https://github.com/node-opcua/node-opcua/pull/1537).

PR #1538, opened 2026-07-03
Implement OPC UA Reverse Connect (Part 6 §7.1.3) on client and server

atx-dgr:
## What & why
Adds spec-compliant OPC UA **Reverse Connect** (Part 6 §7.1.3) to both client and server.
In reverse connect the **server** dials the TCP socket and sends a `ReverseHello` ("RHE");
the client establishes the SecureChannel over it. This is essential when the server sits
behind a firewall/NAT with no open inbound ports (e.g. a factory production network).

Everything above the transport (SecureChannel, Session, Subscriptions) is unchanged — only
who dials the socket and the initial `RHE` differ. All new options are optional; behavior is
unchanged when unused.

## What changed
**Transport** (`node-opcua-transport`)
- new `ReverseHelloMessage` + `decodeReverseHello` (4096-byte field cap); registered `"RHE"`
- `ServerTCP_transport.initReverse` (send RHE, then normal HEL/ACK)
- `ReverseClientTCP_transport` (adopts a server-dialed socket) + `makeReverseClientTransportFactory`

**Server** (`node-opcua-server`, `node-opcua-secure-channel`)
- `ServerSecureChannelLayer.initReverse`; `OPCUAServerEndPoint.createReverseConnection` (outbound dial, reuses channel bookkeeping, connect timeout)
- new `ReverseConnectManager` (dial loop, exponential backoff, redial-on-drop)
- `OPCUAServerOptions.reverseConnect` + start/shutdown wiring

**Client** (`node-opcua-client`)
- `ClientReverseConnect` listener: RHE validation, ServerUri/EndpointUrl matching, DoS guards, durable per-socket error/close handling
- `client.connectReverse(reverseConnect, expectation?)`; secure channel reconciles endpointUrl from the transport; reconnect avoids the outbound cert-fetch dial

**Tests / docs / examples**
- RHE unit tests, `ClientReverseConnect` validation + hardening tests, and an end-to-end test (server dials client → session read)
- `documentation/reverse_connect.md`, two runnable samples, README / supported_features / RELEASE_NOTES entries

## Reviewer notes
- The second commit hardens the client listener socket lifecycle after a max-effort self-review (durable error/close guards to prevent process crashes on peer reset, held-socket eviction, DoS-cap correctness, prompt `stop()`, reverse-aware reconnection).
- The `ReverseConnect` server profile URI is intentionally left gated/commented until cross-stack interop (UA-.NETStandard / open62541) is verified — planned as a follow-up phase.
- Verified: full `node-opcua` build clean; RHE unit + listener + e2e tests green.


erossignon:
Reviewed commits: d62887a8 (feat) and 14a73c47 (hardening).

- **Verdict**: high-quality, spec-aware implementation. The listener lifecycle hardening is genuinely careful and the ServerSecureChannelLayer refactor is a clean DRY win. The one thing holding it back from "production-ready" is test coverage of the riskiest paths — reconnection, the server-side manager, and secured mode are all currently unproven.

### 👍 What's good

- Correct architectural placement. ReverseHelloMessage is hand-authored in node-opcua-transport via buildStructuredType, like the sibling connection-protocol messages 
- Listener hardening is thorough — durable per-socket error/close guards, permanent net.Server error handler, held sockets counted under the DoS cap, unref'd timers, prompt stop().
- Clean DRY refactor in ServerSecureChannelLayer (init and initReverse now share one path via extracted helpers).
- Docs above repo average — handshake diagram, security section, troubleshooting table.
- Backward compatible — everything gated behind optional reverseConnect/connectReverse.


### ⚠️ Needs attention

- Test exhaustiveness (top priority). Zero coverage for the three riskiest areas: end-to-end reconnection/redial-on-drop, the ReverseConnectManager unit behavior (incl. the stop() listener fix this PR makes), and secured (Sign/SignAndEncrypt) mode. Also untested: maxPendingSockets cap, acceptTimeout, listener MAX_RHE_FRAME_LENGTH guard, fragmented-RHE reassembly, endpointUrl matching, server connect-timeout/maxConnections.
waitForConnectionAsync reintroduces the leaked-waiter bug this PR fixes — still exported, non-cancelable, zero internal callers. Remove or make abortable.
Startup backoff inflation — the "endpoint not ready" branch shares the drop-backoff counter (up to ~60s before first dial).
- Three-way endpointUrl reconciliation across transport / client callback / secure-channel layer is fragile.
- Minor: loose structural channel type, duplicated URL-parse snippet, single-endpoint (endpoints[0]) limitation undocumented.

packages/node-opcua-client/source/reverse/client_reverse_connect.ts
```diff
@@ -0,0 +1,387 @@
...
+            cancel: () => {
+                const idx = this.#waiters.indexOf(waiter);
+                if (idx >= 0) {
+                    this.#waiters.splice(idx, 1);
+                }
+            }
+        };
+    }
+
+    public waitForConnectionAsync(expectation?: ReverseConnectExpectation): Promise<IAcceptedReverseConnection> {
```
erossignon: This helper reintroduces the exact bug the hardening commit removed from the client: it's non-cancelable, so a leaked waiter can later hijack an incoming server socket on a disposed transport. It now has zero internal callers (the client uses the cancelable waitForConnection). Please either remove it, or have it return an abortable handle ({ promise, cancel }) and document the hazard — otherwise external users who reach for the obvious async API get the footgun back.

packages/node-opcua-server/source/reverse_connect_manager.ts
```diff
@@ -0,0 +1,192 @@
...
+            state.onAbort = undefined;
+        }
+    }
+
+    #dial(state: ReverseConnectionState): void {
+        if (this.#stopped) {
+            return;
+        }
+        const endpoint = this.#context.getDialEndpoint();
+        if (!endpoint) {
```
erossignon, line 129: The "endpoints not ready yet" retry goes through #scheduleRetry, which doubles currentDelay. On a slow-starting server this inflates the backoff toward maxReconnectDelay (~60s) before the first successful dial. Consider a short fixed retry here (e.g. reconnectDelay without doubling) and only apply exponential backoff to genuine drops.

packages/node-opcua-server/source/reverse_connect_manager.ts
```diff
@@ -0,0 +1,192 @@
...
+                state.retryTimer = undefined;
+            }
+            if (state.pendingSocket) {
+                state.pendingSocket.destroy();
+                state.pendingSocket = undefined;
+            }
+            // established channels are torn down by the endpoint's own shutdown;
+            // just detach OUR OWN re-dial listener so a closing channel does not re-arm.
+            // (removeAllListeners("abort") would also strip the endpoint's _unregisterChannel handler.)
+            if (state.channel && state.onAbort) {
```
erossignon, line 116: Good fix (removing only our own abort listener rather than removeAllListeners). This is subtle and regression-prone — please add a unit test asserting that after stop() the endpoint's own abort listener (_unregisterChannel) is still attached. See suggested RC-MGR-4 below.

packages/node-opcua-server/source/reverse_connect_manager.ts
```diff
@@ -0,0 +1,192 @@
...
+    /** the server ApplicationUri advertised as ReverseHello.ServerUri */
+    getServerUri(): string;
+    /** the SecureChannel EndpointUrl advertised as ReverseHello.EndpointUrl */
+    getEndpointUrl(): string;
+}
+
+interface ReverseConnectionState {
+    clientEndpointUrl: string;
+    pendingSocket?: Socket;
+    channel?: { close(cb: () => void): void; removeListener(event: string, listener: () => void): void };
```
erossignon: This hand-rolled structural type (close/removeListener) drifts easily from the real ServerSecureChannelLayer, and the code calls channel.once(...) which isn't even in the declared shape. The module cycle you're avoiding is with OPCUAServer, not the channel type — consider importing the real channel type here for safety.

packages/node-opcua-client/source/private/client_base_impl.ts
```diff
@@ -630,8 +635,13 @@ export class ClientBaseImpl<Events extends OPCUAClientBaseEvents = OPCUAClientBa
                     }
 
                     if (
-                        err?.message.match("BadCertificateInvalid") ||
-                        err?.message.match(/socket has been disconnected by third party/)
+                        // In reverse-connect mode we must NOT fetchServerCertificate(): that dials the
+                        // server directly (via a temp client), which is impossible for a firewalled
+                        // reverse-connect server and would hang reconnection. Fall through to the plain
+                        // retry, which re-arms the reverse transport and waits for the server to re-dial.
+                        !this._isReverseConnect &&
```
erossignon, line 642: This guard is the crux of reverse-connect reconnection, and it's currently unproven. Please add an e2e that drops an established reverse channel and asserts the client re-accepts the server's redial without ever calling fetchServerCertificate. See suggested RC-E2E-2 below.

packages/node-opcua-client/source/reverse/client_reverse_connect.ts
```diff
@@ -0,0 +1,387 @@
...
+        }
+
+        socket.setNoDelay(true);
+
+        timer = setTimeout(() => {
+            this.#rejectSocket(socket, forget, StatusCodes2.BadTimeout, "no ReverseHello received in time");
+        }, this.#acceptTimeout);
+        timer.unref?.();
+
+        const onData = (chunk: Buffer) => {
```
erossignon, line 279: The chunk reassembly + socket.unshift(leftover) handoff is correctness-critical (bytes after the RHE must survive into the transport) but untested. Please add a test that (a) delivers the RHE split across two TCP writes and (b) appends trailing bytes after the RHE frame and asserts they reach the adopted socket. See suggested RC-LISTEN-7.

packages/node-opcua-client/source/reverse/client_reverse_connect.ts
```diff
@@ -0,0 +1,387 @@
...
+            if (handedOff) {
+                return;
+            }
+            forget();
+        };
+        socket.on("error", onError);
+        socket.on("close", onClose);
+
+        // DoS guard: bound the number of sockets the listener holds (being read + held).
+        if (this.#sockets.size > this.#maxPendingSockets) {
```
erossignon, line 266: The DoS cap is the headline of the hardening commit but has no test. Please add one driving maxPendingSockets + 1 concurrent dials and asserting the last gets an ERR/BadTcpServerTooBusy. See suggested RC-LISTEN-8.

packages/node-opcua-server/source/opcua_server.ts
```diff
@@ -1573,6 +1588,19 @@ export class OPCUAServer extends OPCUABaseServer<OPCUAServerEvents> {
         this.registerServerManager?.start().catch(() => {
             /* empty */
         });
+
+        // start dialing out to any configured reverse-connect clients (OPC UA Part 6 §7.1.3)
+        if (this.options.reverseConnect && this.options.reverseConnect.connections?.length) {
+            this.#reverseConnectManager = new ReverseConnectManager(
+                {
+                    getDialEndpoint: () => this.endpoints[0],
```
erossignon, line 1596: Reverse connect always dials from endpoints[0] and advertises a single getEndpointUrl(), so a multi-endpoint / multi-security-policy server can't select among them via reverse connect. Fine as a limitation, but please note it in documentation/reverse_connect.md.

documentation/reverse_connect.md
```diff
@@ -0,0 +1,125 @@
...
+
+// blocks until an expected server dials in and sends its ReverseHello
+await client.connectReverse(reverseConnect, { serverUri: "urn:the-server:ApplicationUri" });
+
+const session = await client.createSession();
+// ... use the session normally ...
+```
+
+`connectReverse` is "sticky": after a drop it automatically re-accepts the server's next dial-in, so
+you do not need to call it again on reconnection.
```
erossignon, line 108: This advertised behavior is currently unverified by any test — see RC-E2E-2. Worth landing that test alongside the doc claim

erossignon (changes requested):
Suggested tests:

- [ ] RC-E2E-2 — reverse reconnection: establish → force channel drop → assert server redials (respecting reconnectDelay) and the client's session survives, with no direct outbound dial from the client.
- [ ] RC-E2E-3 — secured reverse connect: SignAndEncrypt + serverCertificate supplied up front succeeds; and the negative — omitting the cert fails cleanly (rejects, does not hang).
- [ ] RC-MGR-1 — backoff doubles per failed dial and is capped at maxReconnectDelay; resets to reconnectDelay after a success.
- [ ] RC-MGR-2 — getDialEndpoint() returning undefined retries without inflating the drop-backoff (guards finding #2).
- [ ] RC-MGR-3 — redial fires on channel abort.
- [ ] RC-MGR-4 — after stop(), the endpoint's own abort listener is still attached (guards the removeListener fix).
- [ ] RC-LISTEN-7 — fragmented RHE across two writes, plus trailing bytes after the frame reach the adopted socket.
- [ ] RC-LISTEN-8 — maxPendingSockets + 1 dials → last is rejected with BadTcpServerTooBusy.
- [ ] RC-LISTEN-9 — a socket that connects and sends nothing is closed with BadTimeout after acceptTimeout.
- [ ] RC-LISTEN-10 — a frame whose declared length exceeds MAX_RHE_FRAME_LENGTH → BadTcpMessageTooLarge.
- [ ] RC-LISTEN-11 — endpointUrl-based expectation matches/rejects (only serverUri is exercised today).
- [ ] RC-SRV-1 — createReverseConnection to a host that drops SYNs times out after connectionTimeout (guards the hardening commit's server change); and rejects when maxConnections is reached.

packages/node-opcua-client/source/reverse/client_reverse_connect.ts
atx-dgr: Removed `waitForConnectionAsync()` entirely in 5638874. It had zero internal callers — the client uses the cancelable `waitForConnection()` (stored and cancelled via `_reverseConnectRegistration`), so there's no non-cancelable footgun left on the public API.

packages/node-opcua-server/source/reverse_connect_manager.ts
atx-dgr: Fixed in 5638874. The "endpoints not ready" path now calls `#scheduleRetry(state, /* backoff */ false)`, which schedules a **fixed** `reconnectDelay` and leaves `currentDelay` untouched; exponential backoff is reserved for genuine drops / dial failures.

Guarded by **RC-MGR-2** (3d71119): with `getDialEndpoint()` returning `undefined`, it asserts a constant `reconnectDelay` cadence across the not-ready cycles *and* that the first real backoff still starts at `reconnectDelay` (i.e. `currentDelay` was never inflated).

packages/node-opcua-server/source/reverse_connect_manager.ts
atx-dgr: Added **RC-MGR-4** (3d71119). It attaches a marker `"abort"` listener (standing in for the endpoint's own `_unregisterChannel`) before the channel is handed back, then asserts that after `stop()` the channel still has exactly **one** `"abort"` listener and that it is the endpoint's own — proving only the manager's own re-dial listener was removed.

packages/node-opcua-server/source/reverse_connect_manager.ts
atx-dgr: Fixed in 5638874. `ReverseConnectionState.channel` is now typed as the real `ServerSecureChannelLayer` (`import type { ServerSecureChannelLayer } from "node-opcua-secure-channel"`). `server_end_point.ts` already imports that type, so there's no new `OPCUAServer` cycle, and `channel.once(...)` / `.close(...)` / `.removeListener(...)` now type-check against the actual EventEmitter API.

packages/node-opcua-client/source/private/client_base_impl.ts
atx-dgr: Added **RC-E2E-2** (3d71119). It establishes a reverse channel through a TCP proxy we control, destroys the proxy sockets to drop the channel, and asserts:

1. the server re-dials (respecting `reconnectDelay`) and the client re-accepts it — the session recovers (a read returns `Good` again; the server log shows the session being transferred to the new channel), and
2. the client **never** calls `fetchServerCertificate` (spied via an instance monkey-patch; `callCount === 0`), i.e. no outbound cert-fetch dial on the reverse reconnection path.

packages/node-opcua-client/source/reverse/client_reverse_connect.ts
atx-dgr: Added **RC-LISTEN-7** (3d71119). The RHE is delivered across **three** TCP writes — `< 8` bytes (header incomplete), then header-complete-but-frame-incomplete (listener must buffer and wait), then the remainder — with extra **trailing bytes** appended past the frame boundary. The test then reads those trailing bytes back off the handed-over socket, proving the `socket.pause()` + `socket.unshift(leftover)` handoff preserves post-RHE bytes into the adopted transport.

packages/node-opcua-client/source/reverse/client_reverse_connect.ts
atx-dgr: Added **RC-LISTEN-8** (3d71119). With `maxPendingSockets: 2`, two silent sockets are held (size within the cap) and the third dial pushes the count past the cap and is rejected with a TCP `ERR` carrying `BadTcpServerTooBusy` (the status code is decoded from the frame and asserted).

packages/node-opcua-server/source/opcua_server.ts
atx-dgr: Documented in 5638874. `documentation/reverse_connect.md` now has a **Limitations** section stating that reverse connect always dials from `endpoints[0]` and advertises a single `EndpointUrl`, so a multi-endpoint / multi-security-policy server can't let the client select among them over reverse connect (configure the desired policy on the primary endpoint).

documentation/reverse_connect.md
atx-dgr: This "sticky" reconnection behavior is now verified end-to-end by **RC-E2E-2** (3d71119) — see the reconnection thread on `client_base_impl.ts` above.

atx-dgr:
Thanks for the thorough review — all points addressed. Pushed as 5638874 (review corrections + docs), a748365 (client leak fix), 3d71119 (tests).

### Code corrections
- Fixed non-doubling retry for the "endpoints not ready" startup race (5638874).
- Real `ServerSecureChannelLayer` type instead of the hand-rolled structural type (5638874).
- Removed the non-cancelable `waitForConnectionAsync()` (5638874).
- Documented the single-endpoint limitation (5638874).

### Suggested test checklist — all landed (3d71119)
- **RC-E2E-2** reverse reconnection (redial, session survives, no client outbound dial / `fetchServerCertificate`)
- **RC-E2E-3** secured reverse connect: SignAndEncrypt + `serverCertificate` up front succeeds; omitting it fails cleanly (rejects, no hang)
- **RC-MGR-1** backoff doubles per failed dial, capped at `maxReconnectDelay`, resets after a success
- **RC-MGR-2** `getDialEndpoint()` undefined retries at a fixed delay without inflating the drop-backoff
- **RC-MGR-3** redial fires on channel abort
- **RC-MGR-4** after `stop()` the endpoint's own abort listener is still attached
- **RC-LISTEN-7** fragmented RHE across writes + trailing bytes reach the adopted socket
- **RC-LISTEN-8** `maxPendingSockets + 1` → last rejected with `BadTcpServerTooBusy`
- **RC-LISTEN-9** silent socket closed with `BadTimeout` after `acceptTimeout`
- **RC-LISTEN-10** frame length > `MAX_RHE_FRAME_LENGTH` → `BadTcpMessageTooLarge`
- **RC-LISTEN-11** EndpointUrl-based expectation matches/rejects
- **RC-SRV-1** `createReverseConnection` connect-timeout (host dropping SYNs) and `maxConnections` rejection

Also added, beyond the checklist: RC-MGR-5/6 (idempotent `start`/`stop`; a dial completing after `stop()` closes the channel), RC-LISTEN-12..15 (listener start/stop edge cases, immediate hand-off of an already-held connection), and RHE-7..11 (`decodeReverseHello` error branches + `toString`), lifting patch coverage on the new files to ~97–99%.

### Bonus fix (a748365)
Writing RC-E2E-3's negative case surfaced a real reverse-connect defect: a secured `connectReverse()` with **no** `serverCertificate` threw synchronously deep in the SecureChannel layer *after* `registry.register()` but before the connect callback ran, so the `OPCUAClientImpl` was never unregistered (leaked) and the caller got an opaque empty-message error. Fixed by (a) rejecting early in `connectReverse()` with a clear message, and (b) making the terminal connect-failure handlers idempotently unregister the client — which also closes the same latent leak on the normal `connect()` sync-throw path.

packages/node-opcua-client/source/reverse/client_reverse_connect.ts
erossignon: Confirmed resolved. `waitForConnectionAsync` now has **zero occurrences** anywhere in the tree — the symbol is gone, not merely deprecated, so the non-cancelable waiter can no longer be reached from the public API. Closing.

packages/node-opcua-server/source/reverse_connect_manager.ts
erossignon: Confirmed resolved. `#scheduleRetry(state, backoff = true)` now takes the flag, and the "endpoints not ready" path passes `false`: it schedules a fixed `reconnectDelay` and leaves `currentDelay` untouched, so a slow-starting server no longer inflates the drop backoff toward `maxReconnectDelay`. RC-MGR-2 pins the behaviour. Closing.

packages/node-opcua-server/source/reverse_connect_manager.ts
erossignon: Confirmed resolved. RC-MGR-4 is exactly the test I was asking for: it attaches a marker listener standing in for `_unregisterChannel`, asserts `listenerCount("abort") === 2` after `start()` and `=== 1` after `stop()`, and additionally asserts the survivor is **identity-equal** to the endpoint's own listener. That last assertion is what makes the test regression-proof rather than just counting. Closing.

packages/node-opcua-server/source/reverse_connect_manager.ts
erossignon: Confirmed resolved. `ReverseConnectionState.channel` is now the real `ServerSecureChannelLayer`, brought in as `import type` so no runtime cycle is introduced, and `channel.once(...)` / `.close(...)` are type-checked against the actual class instead of a hand-rolled shape that had already drifted. Closing.

packages/node-opcua-client/source/private/client_base_impl.ts
erossignon: Confirmed resolved. RC-E2E-2 does the real thing rather than a mock: it tunnels the server's dial through a TCP proxy, destroys the proxy sockets to genuinely drop the channel, proves recovery with a `Good` read rather than an internal event name, and asserts `fetchServerCertificateCount === 0`. That last assertion is the crux this guard needed. Closing.

packages/node-opcua-client/source/reverse/client_reverse_connect.ts
erossignon: Confirmed resolved. RC-LISTEN-7 covers both halves of the handoff: the RHE arrives in three writes (fewer than 8 bytes, then header-complete-but-frame-incomplete, then the remainder) to exercise reassembly, and trailing bytes past the frame boundary are asserted to reach the transport, which is the `socket.unshift(leftover)` path. Closing.

packages/node-opcua-client/source/reverse/client_reverse_connect.ts
erossignon: Confirmed resolved. RC-LISTEN-8 drives `maxPendingSockets + 1` concurrent dials and decodes the resulting TCP `ERR` frame to assert the status code is `BadTcpServerTooBusy` — checking the decoded code rather than just "the socket closed" is the right level of strictness here. Closing.

packages/node-opcua-server/source/opcua_server.ts
erossignon: Confirmed resolved. `documentation/reverse_connect.md` now carries a **Limitations** section stating that the server always dials from `endpoints[0]` and advertises a single `EndpointUrl`, so a client cannot select among multiple endpoints or security policies over reverse connect, with the guidance to configure the desired policy on the primary endpoint. Closing.

documentation/reverse_connect.md
erossignon: Confirmed resolved — the documented sticky-reconnection behaviour is now backed by RC-E2E-2, which lands alongside the doc claim as requested. Closing.

erossignon:
## Review round 2 — all previous entries closed, one new item

Thanks for the thorough follow-up. I re-checked all nine open entries **against the code**, not just the replies, and every one is genuinely addressed — I've replied on each thread and resolved them.

What stood out: the tests you added assert the right things rather than the convenient things. RC-MGR-4 checks listener *identity*, not just the count. RC-E2E-2 drops the channel through a real TCP proxy and proves recovery with a `Good` read instead of an internal event name, then asserts `fetchServerCertificateCount === 0`. RC-LISTEN-8 decodes the `ERR` frame to confirm `BadTcpServerTooBusy`. That is the level of rigour this feature needed.

### One remaining item before merge: fixed test ports collide

The four e2e suites bind fixed ports, and seven of them are already claimed by other e2e tests:

| Port | Reverse-connect test | Already used by |
| --- | --- | --- |
| 2231 | `..._basic.ts` | `test_e2e_session_with_X509IdentityToken.ts` |
| 2233 | `..._endpoint.ts` | `test_server_shutdown.ts` |
| 2235 | `..._reconnection.ts` | `test_client_with_null_policyid_in_activate_session.ts` |
| 2237, 2238, 2239, 2240 | `..._secure.ts` | all four allocated elsewhere |

Under the parallel mocha runner this fails intermittently with `EADDRINUSE`, or — worse — silently connects to the wrong server. The `55xx` listener ports were fine; the `22xx` server ports are the whole problem.

I have the fix prepared: reserve **2400–2409** (servers, PKI folder discriminators) and **5600–5609** (client reverse listeners, test proxy) for `test/reverse_connect/`. Both ranges are currently unused across every `.ts`/`.js` source in the repo. It also adds a short `README.md` recording the allocation and the grep to run before adding a port. The two *unit* suites need no reservation — they already listen on port `0` and read back `listeningPort`, which is the pattern to prefer.

This is test-only and touches nothing outside the new directory, so it does not affect the review of the implementation. I could not push it to the head branch — GitHub's "allow edits by maintainers" does not apply to forks owned by an organisation — so I'll land it as a follow-up commit right after merge unless you'd rather cherry-pick it yourself.

Nothing else is blocking. Nice work on this.

erossignon (approved):
All nine review entries verified against the code and resolved. Implementation and test coverage are in good shape.

Merging now; the test-port reallocation follows as a commit on master (see the summary comment above — it could not be pushed to an org-owned fork branch).

erossignon:
Merged as `2d092ac`, and the test-port reallocation landed on master as `0f7a86f`.

Verified before pushing: the whole reverse-connect suite passes on the new ports (7 passing), and it runs cleanly **in parallel** with the tests it used to collide with — no `EADDRINUSE` in either process. (`test_server_shutdown.ts` does fail, but it fails worse standalone than concurrently — a pre-existing timing flake on master, unrelated to this PR.)

Thanks again, David — this was a well-executed piece of work.

erossignon:
released in https://github.com/node-opcua/node-opcua/releases/tag/v2.176.0

PR #1544, opened 2026-07-15
fix(address-space): prevent NodeIdManager symbol-cache collisions and honor AnalogItemType subtypes

erossignon:
## Summary

- `NodeIdManager.constructNodeId()` no longer trusts a symbol-cache hit blindly: it now verifies the cached nodeId is either unoccupied or genuinely owned by the same node (same browseName + same parent) before reusing it. This fixes spurious "already registered" errors when:
  - an inherited and an overriding same-name child collide on the same cache key (e.g. an own-namespace `1:ActualSpeed` next to an inherited `4:ActualSpeed` under the same parent), or
  - sibling nodes under different parents share a truncated cache key (e.g. two `0:EngineeringUnits` properties on sibling AnalogItems), typically during clone operations where the parent chain isn't fully linked yet.

  Genuine duplicates (same browseName, same parent) still resolve to the same id, so `registerNode()`'s duplicate-registration error keeps firing as before.

- `Namespace.addAnalogDataItem()` now honors an explicit `options.typeDefinition` when it is a strict subtype of `AnalogItemType`, instead of unconditionally forcing `AnalogItemType`. The subtype is instantiated so its own mandatory members materialize (e.g. a companion specification's extra property such as PADIM's `LowFlowCutOff`), and the `EURange`/`InstrumentRange`/`EngineeringUnits` handling reuses any properties already cloned by that instantiation instead of creating duplicates.

## Test plan

- [x] `packages/node-opcua-address-space` full test suite passes (1015 passing, 1 pre-existing pending)
- [x] New unit tests in `test/test_nodeid_manager.ts` cover: cross-namespace same-local-name siblings, colliding-key-with-different-occupant (different name, different parent), genuine-duplicate-still-throws, and the no-parent (top-level node) comparison branches
- [x] New unit tests in `test/data_access/subtest_analog_item_type.ts` cover: subtype `typeDefinition` honored + mandatory member materialized, `typeDefinition` passed as `NodeId` / browseName string, invalid (non-AnalogItemType-subtype) `typeDefinition` throws, and subtype used without `instrumentRange`/`engineeringUnits` options
- [x] Full monorepo `tsc -b` build passes with no type errors

### DIFF
--- a/packages/node-opcua-address-space/source/namespace_data_access.ts
+++ b/packages/node-opcua-address-space/source/namespace_data_access.ts
@@ -2,7 +2,8 @@ import type {
     AddVariableOptionsWithoutValue,
     AddYArrayItemOptions,
     BindVariableOptions,
-    UADataType
+    UADataType,
+    UAVariableType
 } from "node-opcua-address-space-base";
 import type { NodeIdLike } from "node-opcua-nodeid";
 import type { UAAnalogItem, UADataItem } from "node-opcua-nodeset-ua";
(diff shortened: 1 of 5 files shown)


erossignon:
Addressed in 9458ba848: compare `browseName.namespaceIndex` literally instead of normalizing 0 → own namespace (an explicit ns=0 could otherwise mask a real cross-namespace collision), and fixed 5 unit tests that were injecting the forced cache collision under the wrong key (missing the parent-name prefix), so they weren't actually exercising the collision path.

PR #1555, opened 2026-08-10
fix(server-configuration): give each TrustList a real, unique backing file

erossignon:
## Symptom

Browsing a server's `ServerConfiguration` TrustList with UaExpert logs an
ENOENT against the in-memory backing file and answers `Bad_UnexpectedError`:

```
:file_type_helpers :426  ENOENT: no such file or directory, open '/tmpFile3'
    at _Superblock.getResolvedLinkOrThrow (...)
    at _Superblock.openFile (...)
    at Volume.open (...)
```

## Two independent defects behind it

`promoteTrustList` did:

```ts
let counter = 0;                        // module scope
const filename = `/tmpFile${counter}`;
counter += 1;
installFileType(trustList, { filename, fileSystem: MemFs });
```

**1. The backing file is never created.** It only appears when `_openTrustList`
runs `writeTrustList(...)`, inside the `Open` handler, and only on the branch
where the group has a `$$certificateManager`. Until then the path dangles, and
two readers hit it:

- `file_type_helpers.ts:106`, the `Size` binding calls `statSync(filename)`.
  Simply browsing the TrustList reads `Size`.
- `file_type_helpers.ts:405`, `openFile` calls `abstractFs.open(filename, flags)`
  when `Open` falls through to the raw implementation, which is what happens for
  a certificate group that was promoted without a certificate manager.

An empty file at promotion time makes a group with nothing to serve read as an
empty TrustList, which is the correct answer, rather than an error.

**2. The filename is not unique.** `counter` is unique only within one
*instance* of this module, while `memfs` exports a single process-wide volume.
A process that loads two copies of `node-opcua-server-configuration` (duplicated
transitive dependency, or a monorepo resolving two versions) gives two different
TrustLists the same `/tmpFile0` on the same volume, and they silently overwrite
each other's content on every `Open`.

A NodeId alone does not fix this either: two `OPCUAServer` instances in one
process have separate address spaces containing the same TrustList NodeId. The
path now combines a per-address-space id with the NodeId, and the id is stored
on the address space object rather than in a module-scoped map so that every
duplicate copy of the module observes the same value.

## Tests

`packages/node-opcua-server-configuration`: 133 passing, including
`test_trust_list_file_lock.ts` and the full `test/server/` set. Package and test
tsconfigs both typecheck clean.

### DIFF
--- a/packages/node-opcua-server-configuration/source/server/promote_trust_list.ts
+++ b/packages/node-opcua-server-configuration/source/server/promote_trust_list.ts
@@ -643,6 +643,29 @@ export async function promoteTrustList(trustList: UATrustList) {
         MemFs.writeFileSync(filename, Buffer.alloc(0));
     }
 
+    // `memfs` exports one volume for the whole process and nothing here ever
+    // released these paths. Each Open rewrites the full serialized TrustList
+    // (every trusted cert, issuer cert and CRL) and it then stays resident
+    // for the lifetime of the process. That was masked while the filename
+    // came from a module counter, because a second address space reused
+    // `/tmpFile0` and overwrote the first one's bytes; now that the paths are
+    // distinct, a process that builds and tears down servers repeatedly (test
+    // suites, an in-process restart) would accumulate one blob per TrustList
+    // per server. Release them with the address space.
+    //
+    // Registering once per TrustList is bounded: the $$promoted guard above
+    // means this runs once per node, and registerShutdownTask has no
+    // unregister counterpart.
+    trustList.addressSpace.registerShutdownTask(() => {
+        try {
+            if (MemFs.existsSync(filename)) {
+                MemFs.unlinkSync(filename);
+            }
+        } catch (err) {
+            debugLog("could not release trust list backing file", filename, (err as Error).message);
+        }
+    });
+
     // Store filename for use in _closeAndUpdate
     trustListEx.$$filename = filename;
     // Initialize write lock flag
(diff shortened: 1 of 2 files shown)


erossignon:
## Reproduced, and a memory-lifetime follow-up

**Root cause confirmed.** `TLBF-1` promotes a certificate group with no `certificateManager` and opens its TrustList. Against the pre-fix source it fails with exactly the reported signature:

```
Error: cannot open file statusCode = BadUnexpectedError (0x80010000) mode = Read
  ENOENT: no such file or directory, open '/tmpFile3'
      at Superblock.openFile ...
      at Volume.Superblock.open ...
```

With the fix it returns a handle on an empty TrustList.

**MemFs lifetime.** Reviewing the change surfaced a leak it would have made worse. `memfs` exports one volume for the whole process, every `Open` rewrites the full serialized TrustList (all trusted certs, issuer certs and CRLs) into it, and nothing ever unlinked those paths. The module counter had been masking it: a second address space reused `/tmpFile0` and overwrote the first one's bytes, so residency stayed bounded by accident, at the cost of the correctness bug this PR fixes. With distinct paths, a process that builds and tears down servers (test suites, in-process restart) would retain one blob per TrustList per server.

The paths are now released through `addressSpace.registerShutdownTask`. Registration is bounded: the `$$promoted` guard means it runs once per node, which matters because `registerShutdownTask` has no unregister counterpart.

**Coverage**

- `TLBF-1` unbound group can be opened
- `TLBF-2` groups do not share a backing path. It cannot exercise the duplicate-module case, which needs two copies loaded at once, so it only guards a regression to one shared name
- `TLBF-3` backing files are gone after `addressSpace.shutdown()`

136 passing in `node-opcua-server-configuration`; source and test tsconfigs both clean.

PR #1564, opened 2026-08-19
fix: don't abandon nodeset post-loading when one task throws

erossignon:
## Problem

Loading a nodeset that contains a single malformed `<Value>` silently leaves the whole address space half-initialised.

Real-world trigger shipped in `node-opcua-nodesets`: `Opc.Ua.PADIM.NodeSet2.xml` has two `ValveSwitchingCyclesCounter` variables (`ns=3;i=6016`, `ns=3;i=6031`) declared `DataType="UInt32" ValueRank="1"` whose value is a `<uax:ListOfInt32>`. `verifyVariantCompatibility` rightly rejects the Int32/UInt32 mismatch, and then:

1. `performPostLoadingTasks` (`load_nodeset2.ts`) caught the error, logged it… and **re-ran the same task inside the `catch`**. The second throw escaped the loop.
2. `finalSteps()` was abandoned right there — before `ensureDatatypeExtracted()`, before `postTasks1_InitializeVariable`, `postTasks2_AssignedExtensionObjectToDataValue` and `promoteObjectsAndVariables()`.
3. The outer `catch (err) { renderError(err); }` swallowed everything: `renderError` in `node-opcua-assert` is literally `return err;`. `generateAddressSpace()` resolved normally.
4. Any later use of a structured DataType — e.g. `namespace.addAnalogDataItem({ engineeringUnitsRange })`, or instantiating any PADIM `AnalogSignalVariableType` subtype (`EURange`) — died in `AddressSpace.getExtensionObjectConstructor` with the opaque
   `TypeError: Cannot read properties of undefined (reading 'getExtensionObjectConstructorFromDataType')`
   because it read `$$extraDataTypeManager` directly instead of going through the guarded `getDataTypeManager()` accessor.

Minimal repro on `master` (no custom nodeset needed):

```ts
const a = AddressSpace.create();
a.registerNamespace("urn:test");
await generateAddressSpace(a, [nodesets.standard, nodesets.di, nodesets.padim, nodesets.irdi], {});
console.log(!!(a as any).$$extraDataTypeManager);          // false  (true with [standard, di])
a.getOwnNamespace().addAnalogDataItem({ browseName: "T", dataType: "Float", engineeringUnitsRange: { low: 0, high: 100 } });
// TypeError: Cannot read properties of undefined (reading 'getExtensionObjectConstructorFromDataType')
```

## Fix

- **`load_nodeset2.ts` — `performPostLoadingTasks(stage, tasks)`**: a task that throws is logged once (`[NODE-OPCUA-W36] … post-loading task failed during "<stage>" and was skipped: <message>`) and skipped. No more retry-in-catch, so one bad `<Value>` can no longer abort the rest of the pipeline. Each stage is labelled at the call site so the log says *which* stage failed.
- **`load_nodeset2.ts` — outer `finalSteps()` catch**: log `[NODE-OPCUA-E30] … final post-loading steps failed: <message>` and **re-throw**, so `generateAddressSpace()` rejects instead of handing back a half-initialised address space. (`renderError` import removed — it was a no-op.)
- **`address_space.ts` — `getExtensionObjectConstructor`**: explicit error when `$$extraDataTypeManager` is missing, naming the DataType and pointing at `generateAddressSpace()` / `ensureDatatypeExtracted()`, instead of a `TypeError` on `undefined`.

## Tests

`test/test_load_nodeset2_resilience.ts` + fixture `test_helpers/test_fixtures/nodeset_with_mismatched_value_datatype.xml` (a UInt32 array variable with a `ListOfInt32` value next to a good UInt32 variable):

- RES-1: loading succeeds, the good variable is initialised (`42`, `Good`), the bad one is not `Good`, and the DataTypeManager exists.
- RES-2: `addAnalogDataItem({ engineeringUnitsRange })` works after loading such a nodeset (the exact downstream failure).
- RES-3: `getExtensionObjectConstructor` raises the explicit message when the manager is missing.

## Not in this PR

The PADIM nodeset data issue itself (`ListOfInt32` for a `UInt32` array). Arguably the XML variant reader could coerce compatible integer lists to the declared DataType; that's a separate, broader change. With this PR those two counters simply stay uninitialised (logged as W36) and everything else works.

### DIFF
--- a/packages/node-opcua-address-space/source/loader/load_nodeset2.ts
+++ b/packages/node-opcua-address-space/source/loader/load_nodeset2.ts
@@ -16,7 +16,7 @@ import type {
     UAVariable,
     UAVariableType
 } from "node-opcua-address-space-base";
-import { assert, renderError } from "node-opcua-assert";
+import { assert } from "node-opcua-assert";
 import { coerceBoolean, coerceByte, coerceInt32, StatusCodes } from "node-opcua-basic-types";
 import { DataTypeIds } from "node-opcua-constants";
 import {
@@ -1160,7 +1160,19 @@ function makeNodeSetParserEngine(addressSpace: IAddressSpace, options: NodeSetLo
                     chalk.green("DONE")
             );
 
-        async function performPostLoadingTasks(tasks: Task[]): Promise<void> {
+        /**
+         * run every queued post-loading task.
+         *
+         * A task that throws (typically: a `<Value>` in a third-party nodeset that does not
+         * match the declared DataType) is logged and skipped — it must NOT abort the remaining
+         * tasks nor the remaining stages of finalSteps(). Previously the failing task was
+         * re-run inside the catch block, the second throw escaped the loop, finalSteps() was
+         * abandoned and `ensureDatatypeExtracted` was never called: the address space was
+         * left half-initialised (no DataTypeManager) and any later use of a structured
+         * DataType failed with an opaque
+         * "Cannot read properties of undefined (reading 'getExtensionObjectConstructorFromDataType')".
+         */
+        async function performPostLoadingTasks(stage: string, tasks: Task[]): Promise<void> {
             for (const task of tasks) {
                 try {
                     await task(addressSpace1);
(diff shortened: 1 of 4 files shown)


erossignon:
The 5 failing CI jobs all fail on the exact same test — `PushCertificateManager updateCertificate should store multiple issuer certificates correctly` (`node-opcua-server-configuration/test/server/test_push_certificate_manager_server_impl.ts:522`, `AssertionError: expected 0 to equal 2`).

This is not caused by this PR: the same test fails with the identical assertion on `master`'s own CI run at the commit this branch is based on (54765d08b), before this diff existed — https://github.com/node-opcua/node-opcua/actions/runs/32247417419/job/96051132666. This PR only touches `node-opcua-address-space` (nodeset post-loading + `getExtensionObjectConstructor`) and adds a new, fully passing test file there; there's no code path connecting it to certificate/issuer storage. Looks like a pre-existing, timing-sensitive flake unrelated to this change.

erossignon:
Root cause identified and fixed at the source (this was a real, deterministic bug — not a flake — in `node-opcua-server-configuration`'s certificate/private-key comparison, unrelated to this PR's address-space change): #1565

PR #1622, opened 2026-09-03
fix(server-configuration): relay out-of-band TrustList changes as applyChangesCompleted

erossignon:
## Summary
- `installPushCertificateManagementOnServer` now relays a TrustList change to `applyChangesCompleted` when it comes from anywhere other than a live `ApplyChanges` RPC (an admin script, a Pull-model client) writing into the trusted-certs folder `CertificateManager` already watches. Previously only the RPC path fired that event, so a server relying on it to leave `NoConfiguration` once provisioned could stay stuck there forever if trust was populated any other way.

## Test plan
- [x] `packages/node-opcua-server-configuration/test/server/test_apply_changes_completed_relay.ts` (new): fires for a directly-trusted certificate, does not fire for a rejected one
- [x] `test_no_configuration_certificate_relaxation.ts`, `test_server_with_push_certificate_management.ts`: no regressions
- [x] typecheck, biome, check:importext, check:shortcircuit

### DIFF
--- a/packages/node-opcua-server-configuration/source/server/install_push_certificate_management.ts
+++ b/packages/node-opcua-server-configuration/source/server/install_push_certificate_management.ts
@@ -315,6 +315,58 @@ export async function installPushCertificateManagementOnServer(
         });
     });
 
+    // ── Report an out-of-band TrustList change as trustListUpdated ─────
+    //
+    // The TrustList server methods (AddCertificate, CloseAndUpdate) are not
+    // the only way a server's TrustList can go from empty to populated: an
+    // administrator, a provisioning script, or a Pull-model client can
+    // write a certificate straight into serverCertificateManager's
+    // trusted-certs folder. Only the method path emits "trustListUpdated",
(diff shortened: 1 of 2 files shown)


erossignon:
On hold. The CTT-side fix is being done without this PR: the server's own CertificateManager already exposes `certificateAdded`, and `isTrustListEmpty()` is correct synchronously, so the consumer can observe the change directly without an SDK change.

Also recording what this PR cost, in case it is revived: emitting `applyChangesCompleted` for a trust-list change broke `CMZ-1` in `test_certificate_expired_alarm.ts` (3/3 runs red), because that event carries channel teardown and endpoint suspend/resume with it. Switching to `trustListUpdated` fixed it (5/5 green). The current head has that correction.

The genuinely reportable issue found along the way is narrower and belongs in node-opcua-pki rather than here: each CertificateManager registers five non-persistent chokidar watchers, and in a ~4000-test run the `certificateAdded` event stopped arriving within 10s on Linux CI while firing in ~40ms locally.

PR #1625, opened 2026-09-05
fix(server): a UserNameIdentityToken without the serverNonce is invalid

erossignon:
## Problem

OPC 10000-4, *Token Encryption and Proof of Possession*, requires a client to append the session's `serverNonce` to the password before encrypting a `UserNameIdentityToken`. The server correctly refused a token that omitted the nonce, but answered **`BadUserAccessDenied`** — which tells the client its *credentials* were refused and invites a retry — instead of **`BadIdentityTokenInvalid`**, which names the actual fault: a malformed token the client must fix.

CTT *Security / Security User Name Password / 006* exercises exactly this (insecure channel, valid username, empty password, nonce not appended) and reports:

```
RECOMMENDATION: The server returned 'BadUserAccessDenied (0x801f0000)' when sending an
empty password without nonce which is acceptable. According to the specification
'Part 4, Token Encryption and Proof of Possession' the server should return
Bad_IdentityTokenInvalid instead.
```

## Cause

`extractPasswordFromDecryptedBlob` already verified the trailing nonce, but the only channel back to the caller was a boolean:

```
userNameIdentityTokenAuthenticateUser  ->  callback(null, false)
isUserAuthorized                       ->  callback(null, false)
_on_ActivateSessionRequest             ->  rejectConnection(BadUserAccessDenied)
```

`!authorized` was hard-mapped to `BadUserAccessDenied`, so a malformed blob and a wrong password were indistinguishable by the time the status code was chosen.

## Change

The authorization callback gains an **optional third argument** carrying the precise rejection code:

- nonce missing / mismatched / blob malformed -> `BadIdentityTokenInvalid`
- blob that will not decrypt at all -> the same code (equally a malformed token, not a credential failure)
- `_on_ActivateSessionRequest` -> `rejectConnection(this, statusCode || StatusCodes.BadUserAccessDenied)`

The `|| BadUserAccessDenied` fallback keeps this non-breaking: `userNameIdentityTokenAuthenticateUser` and `isUserAuthorized` are `protected`, so any subclass override still calling `callback(null, false)` keeps today's behaviour. A genuinely wrong password — which reaches `userManager.isValidUser` only *after* the nonce check passes — still yields `BadUserAccessDenied`.

## Test

`test_e2e_username_identity_token_missing_nonce.ts` reproduces the CTT scenario against a real client/server pair: channel security `None`, server offering `None + Basic256Sha256` so the username policy is an encrypted one. It intercepts `performMessageTransaction` and re-encrypts the `ActivateSessionRequest` blob as `length | password` with **no trailing nonce** — using the *correct* password, so the token is proven to be rejected on its shape rather than on its credentials.

Verified both ways:

| | result |
|---|---|
| with the fix | passes |
| fix reverted, server rebuilt | fails: `expected 'BadUserAccessDenied (0x801f0000)' to match /BadIdentityTokenInvalid/` |

The existing `test_e2e_secure_UserNameIdentityToken.ts` suite (8 tests, including two that assert `BadUserAccessDenied` for wrong credentials) and the `extractPasswordFromDecryptedBlob` unit tests (7) still pass. `biome check`, `check-test-ports`, `check-import-extension` and `check-test-types` are clean.

### DIFF
--- a/packages/node-opcua-end2end-test/test/end_to_end/test_e2e_985.ts
+++ b/packages/node-opcua-end2end-test/test/end_to_end/test_e2e_985.ts
@@ -346,7 +346,10 @@ describe("test reconnection when server stops and change it privateKey and certi
             await server.shutdown();
         }
         should.exist(_capturedError);
-        should(_capturedError?.message).match(/BadUserAccessDenied|BadIdentityTokenRejected/);
+        // the ciphertext is corrupted, so the blob either fails to decrypt or decrypts to
+        // garbage that carries no serverNonce: either way the token is malformed, and the
+        // server names that rather than claiming the credentials were refused.
+        should(_capturedError?.message).match(/BadIdentityTokenInvalid/);
         hacked.should.eql(true);
     });
     it("T6 - server shall not crash if the user identity token is corrupted ActivateSession", async () => {
--- a/packages/node-opcua-end2end-test/test/secure_connection/test_e2e_username_identity_token_missing_nonce.ts
+++ b/packages/node-opcua-end2end-test/test/secure_connection/test_e2e_username_identity_token_missing_nonce.ts
@@ -1,0 +1,123 @@
+import { createPublicKey } from "node:crypto";
+import "should";
+import {
+    ActivateSessionRequest,
+    getCryptoFactory,
+    MessageSecurityMode,
+    OPCUAClient,
+    OPCUAServer,
+    type Request,
+    SecurityPolicy,
+    UserNameIdentityToken,
+    UserTokenType
+} from "node-opcua";
+import { extractPublicKeyFromCertificateSync, toPem } from "node-opcua-crypto";
+import { describeWithLeakDetector as describe } from "node-opcua-leak-detector";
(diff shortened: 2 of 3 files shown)


erossignon:
CI caught one more site: `test_e2e_985.ts` **T5** ("server shall not crash if password is wrongly encrypted by client in ActivateSession") corrupts the first three bytes of the encrypted password blob, so it went through the *decrypt-failure* branch and asserted `/BadUserAccessDenied|BadIdentityTokenRejected/`.

That case is the same defect this PR fixes: the ciphertext either fails to decrypt or decrypts to garbage carrying no `serverNonce`, so the token is malformed and the credentials are never reached. `BadIdentityTokenInvalid` is the honest answer for both paths, so the assertion is updated rather than the server behaviour narrowed. The outcome is deterministic, so it now names one code instead of two.

All four failing jobs (`build 22.x/24.x`, `build_on_windows`, `build_per_package`, `coverage`) failed on that single assertion and nothing else. Locally, `test_e2e_985.ts` now passes 6/6.

PR #1626, opened 2026-09-05
fix(server): size a PublishResponse to the channel, not to a count

erossignon:
## The bug

A `PublishResponse` is assembled by **count** — `maxNotificationsPerPublish`, capped by the static `Subscription.maxNotificationPerPublishHighLimit` (1000) — and the encoded size is never consulted:

```js
const m = this.maxNotificationsPerPublish;
while (i < m && this._pending_notifications.size > 0) { ... }
```

A count cannot express a byte limit. A Boolean notification and one carrying a 64 kB ByteString differ by four orders of magnitude, so **no constant is right for both**. A client that declared a modest `maxMessageSize` when it opened the channel then receives a message it refuses.

## How it showed up

Running the OPC Foundation CTT (1.05.513) against a server advertising `maxMonitoredItemsPerSubscription: 5000`. `Subscription Basic/initialize.js` creates `MaxSupportedMonitoredItems` items, then a single `Publish` carries their initial values:

```
Subscription Basic / Err-019.js
  ERROR: Publish.Response.ResponseHeader.ServiceResult is Bad: BadTcpMessageTooLarge (0x80800000)
  ERROR: Publish() failed, status: Good (0x00000000)
```

Note the two lines together: the **service** result is `Good` — the server answered normally — while the response header carries a transport error. That only happens when the rejection is client-side, *after* the server has replied. `BadTcpMessageTooLarge` appears nowhere in this repository's server code.

Eight scripts were affected: `Subscription Basic` 049 and Err-019…Err-024, and `Monitor Basic` 010.

## Why tuning the constant cannot fix it

| `maxNotificationPerPublishHighLimit` | result |
|---|---|
| 10000 | initial-value burst refused — `BadTcpMessageTooLarge` ×8 |
| 1000 (default) | still refused, same 8 scripts |
| 500 | still refused, **and** `Base Info ResendData Method` degrades ×7 |

Lowering it far enough to fit breaks the units that need a burst delivered inside one publishing interval; raising it makes the messages worse. The two constraints do not meet at any constant, because the bound is bytes and the knob is items.

## The change

The budget comes from the `maxMessageSize` the client declared in its Hello, reached through the session already hanging off the subscription — no interface widening was needed:

```ts
this.$session?.channel?.getTransportSettings?.()?.maxMessageSize
```

`0` means no limit was announced, and behaviour is then exactly as before. Bytes accumulate alongside the count; whichever bound is reached first ends the message and the remainder goes out under `moreNotifications`.

Two cases were decided deliberately rather than left to be discovered:

- **Progress is guaranteed.** The first notification of a message is always taken, however heavy, so an outsized value *leads the next message* instead of being deferred behind an empty one. Without that guard this change can livelock — a subscription emitting nothing while `moreNotifications` stays true. `PRS-5` pins it.
- **A value too large to fit even alone** can never reach that client. Sending it anyway makes the transport tear down the channel, taking every other subscription on it; dropping it leaves the client waiting for a value that never comes. It is delivered with its `clientHandle` and timestamps intact but carrying `BadResponseTooLarge` — *"the response message size exceeds limits set by the client or server"* — and no value. The item recovers by itself when its value next fits, the status change being a change like any other.

## Tests

`packages/node-opcua-server/test/test_publish_response_sizing.ts` — deliberately abstract: no socket, no channel, no address space, no sampling. Notifications go straight onto the pending queue and the assembled message is weighed with `binaryStoreSize()`, because bytes are the unit the client rejects on. Runs in ~50 ms.

| | |
|---|---|
| PRS-1 | splits on bytes even when the item count is far below the cap |
| PRS-2 | never emits a response larger than the negotiated `maxMessageSize` |
| PRS-3 | sends an outsized value alone rather than dropping or deferring it |
| PRS-4 | reports `BadResponseTooLarge` for a value that cannot fit at all |
| PRS-5 | always makes progress — never an empty notification with `moreNotifications` set |
| PRS-6 | a degraded item reports normally again once its value fits |

Before the change PRS-1 fails with `one message carried 164393 bytes, more than the 65536 the client negotiated` — from twenty items, four hundred times below the count cap, which is the point: the cap never engages.

## Not yet verified

Only these six tests have been run. The change touches the core publish path, so **the full `node-opcua-server` suite has not been exercised** — I am relying on CI for that, and expect it to be the real gate on this PR. The `notificationMessageOverhead` reserve (2048 bytes) is a deliberate over-estimate rather than a computed header size; erring high costs one extra message, erring low costs the whole message.

### DIFF
--- a/packages/node-opcua-server/source/server_subscription.ts
+++ b/packages/node-opcua-server/source/server_subscription.ts
@@ -20,7 +20,8 @@ import { assert } from "node-opcua-assert";
 import type { Byte, UInt32 } from "node-opcua-basic-types";
 import { SubscriptionDiagnosticsDataType } from "node-opcua-common";
 import { AttributeIds, isValidDataEncoding, NodeClass, type QualifiedNameLike } from "node-opcua-data-model";
-import type { DataValue, TimestampsToReturn } from "node-opcua-data-value";
+import type { TimestampsToReturn } from "node-opcua-data-value";
+import { DataValue } from "node-opcua-data-value";
 import { checkDebugFlag, make_debugLog, make_warningLog } from "node-opcua-debug";
 import { NodeId } from "node-opcua-nodeid";
 import type { NumericRange } from "node-opcua-numeric-range";
@@ -60,6 +61,14 @@ const debugLog = make_debugLog("server_subscription");
 const doDebug = checkDebugFlag("server_subscription");
 const warningLog = make_warningLog("server_subscription");
 const maxNotificationMessagesInQueue = 100;
+
+/**
+ * Bytes reserved for everything in a PublishResponse that is not a notification:
+ * the response and notification-message headers, the DataChangeNotification
+ * wrapper and the diagnostic-info arrays. Deliberately generous - overshooting
+ * the budget costs one extra message, exceeding it costs the whole message.
+ */
+const notificationMessageOverhead = 2048;
 
 export interface SubscriptionDiagnosticsDataTypePriv extends SubscriptionDiagnosticsDataType {
     $subscription: Subscription;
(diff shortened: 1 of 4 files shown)


erossignon:
## Correction on the second commit's rationale

The transport commit says the 128 kB floor is "very likely the actual mechanism" behind the `BadTcpMessageTooLarge` seen with the CTT. **That is not established, and I should not have written it that way.**

I inferred it from reading `clamp_value` and never confirmed what the CTT actually declares in its Hello. Having since enabled `NODEOPCUADEBUG="TRANSPORT{HELACK}"` on the server and driven the failing CTT unit against it, I could not capture the CTT's inbound Hello at all — the only traced Hello was from a node-opcua client (`maxMessageSize = 0`, server adopts its 16 MB default), and the other limit values in the log belong to the server's own outbound registration client and the remote server's Acknowledge.

So the position on each commit is:

- **`fix(server): size a PublishResponse to the channel`** — the defect is demonstrated: batching by count cannot express a byte bound, PRS-1 shows 164 kB emitted from twenty items, four hundred times below the count cap.
- **`fix(transport): never raise a MaxMessageSize`** — correct on the specification alone. OPC 10000-6 §7.1.2.3 makes Hello's `MaxMessageSize` the largest response the *client* will accept, and a floor that raises it lets the server exceed a stated ceiling. That reasoning stands on its own. Whether it explains the CTT observation is **unproven**, and the commit message overstates it.

Both changes are still worth having; the second is a spec-conformance fix rather than a confirmed root cause. Happy to reword or split it if a maintainer prefers.

